Issue #5, opened 2017-06-22
Columns with spaces cause BulkInsertOrUpdate to fail


gm112:
**How to reproduce: ** Create a table that has a column name with spaces in it. Then, try to run a query.

This can actually be easily fixed by detecting whitespace in the Column names and wrapping them with brackets [] when whitespace is detected. 

NOTE: This does not occur on a mere BulkInsert, but a BulkInsertOrUpdate call.

borisdj:
I have just updated repository with this fix.
Nuget library is not yet published, but you can test it from code first...

gm112:
I can confirm that your fixed seems to work! Nice job

borisdj:
Nuget 1.0.4 published, closing this issue.

borisdj:
NuGet 1.0.4 published, closing this issue.

PR #6, opened 2017-06-22
Fix BulkInsertOrUpdate on whitespaced columns

gm112:
Contain the columnName within brackets[] to allow for column names with spaces in them.

### DIFF
--- a/EFCore.BulkExtensions/SqlQueryBuilder.cs
+++ b/EFCore.BulkExtensions/SqlQueryBuilder.cs
@@ -71,8 +71,8 @@ namespace EFCore.BulkExtensions
             string commaSeparatedColumns = "";
             foreach (var columnName in columnsNames)
             {
-                commaSeparatedColumns += prefixTable != null ? $"{prefixTable}.{columnName}" : columnName;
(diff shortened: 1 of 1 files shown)


gm112:
This addresses #5

borisdj:
I've already made that fix before you have sent merge request so I am not able to merge this now.
Additionally changes are done not only in _SqlQueryBuilder.cs_ but also in _SqlBulkOperation.cs_ and _TableInfo.cs_ so that _TableId_ would be in brackets as well.
And Tests are updated to follow new syntax.
Please check the last version from source, not nuget because nuget is not published yet, and see if it works for you.

gm112:
Yeah - my apologies I probably should have gone through the other files a bit more thoroughly. But that is fantastic you managed to workout the problem!

And on that note, I was unable to reproduce the exception using the source build from today in both the Unit Tests and in my project. :) Thank you!

borisdj:
Nice,
I will now close this merge, and once I update the NuGet as well I will then close the original issue.

borisdj:
Close merge.

Issue #15, opened 2017-08-11
Class structures with Discriminator aren't supported


Alexx999:
Trying to do BulkInsert with a structure that maps several types to one table using some Discriminator field results in the IndexOutOfRange exception in FastMember.
This is obviously due to fact that Discriminator isn't present in instances.

borisdj:
You can bypass this problem by making custom Discriminator, which you probably have do to anyway in order to set value of Discriminator column, so you need real Discriminator property.
Example:
```
public abstract class Person
{
    public int Id { get; set; }

    public string Discriminator { get; protected set; }
	
    ...
}

public class Instructor : Person
{
    public Instructor()
    {
        Discriminator = nameof(Instructor);
    }
	
    ...
}

public class Student : Person
{
    public Student()
    {
        Discriminator = nameof(Student);
    }

    ...
}
```

And in DbContext.cx
```
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
    modelBuilder.Entity<Person>().HasDiscriminator<string>("Discriminator")
        .HasValue<Instructor>(nameof(Instructor))
        .HasValue<Student>(nameof(Student));
}
```

Alexx999:
Well, problem is that I don't need to set any values there - EF Core is quite happily doing that for me and saving that into DB. However, it's impossible to save such entities with BulkExtension as there's no shadow property support.

borisdj:
Configuration from the example above Does work with BulkExtension since here Discriminator is real property and not shadow, so there is no issue with FastMember either.

Table-Per-Hierarchy inheritance model can be set in 2 ways.
One is when EF does it automatically by Convention in which case you have shadow Property.
And second is like in this example where you explicitly define Discriminator property in Entity and configure it with _.HasDiscriminator()_.

Alexx999:
I wonder if there are more cases where EF Core uses shadow properties. At least they can be freely added by a user so having some support for them (at least throwing NotSupportedException to let user know that they're not supported) is clearly required.
Because now it requires some debugging to figure out why the hell objects are not saving which isn't cool.

borisdj:
I agree that Shadow support is good and I will integrate it, just to figure the cleanest way.
Comment was to add info that TPH can be done with explicit config so without Shadow properties.

Alexx999:
Yeah, probably my initial message isn't very clear - I know that you can do that explicitly, however, by default EF Core will do that implicitly if your PK configuration allows that.

[...]

PR #17, opened 2017-08-11
Add simple shadow property support

Alexx999:
This is my take on issue #15 - it's not perfect as it potentially can decrease performance for types without shadow properties (probably can add "if" to ObjectReaderEx.Create and return ObjectReader if no shadow properties present).
Also, it uses reflection to get current item out of ObjectReader. The only other option I see is copying entire ObjectReader implementation onto our project to make "current" field protected instead of private which isn't great either.

### DIFF
--- a/EFCore.BulkExtensions/ObjectReaderEx.cs
+++ b/EFCore.BulkExtensions/ObjectReaderEx.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Reflection;
+using FastMember;
+using Microsoft.EntityFrameworkCore;
+
+namespace EFCore.BulkExtensions
+{
+    internal class ObjectReaderEx : ObjectReader
+    {
+        private readonly HashSet<string> _shadowProperties;
+        private readonly DbContext _context;
+        private string[] _members;
+        private static TypeAccessor _accessor = TypeAccessor.Create(typeof(ObjectReader), true);
+        private FieldInfo _current;
+
+        internal static object GetInstanceField(Type type, object instance, string fieldName)
+        {
+            FieldInfo field = type.GetField("current", BindingFlags.Instance | BindingFlags.NonPublic);
+            return field.GetValue(instance);
+        }
+
+        public ObjectReaderEx(Type type, IEnumerable source, HashSet<string> shadowProperties, DbContext context, params string[] members) : base(type, source, members)
+        {
+            _shadowProperties = shadowProperties;
+            _context = context;
+            _members = members;
+            _current = typeof(ObjectReader).GetField("current", BindingFlags.Instance | BindingFlags.NonPublic);
+        }
+
+        public static ObjectReaderEx Create<T>(IEnumerable<T> source, HashSet<string> shadowProperties, DbContext context, params string[] members)
+        {
+            return new ObjectReaderEx(typeof(T), source, shadowProperties, context, members);
+        }
+
+        public override object this[string name]
+        {
+            get
+            {
+                if (_shadowProperties.Contains(name))
+                {
+                    var current = _current.GetValue(this);
+                    return _context.Entry(current).Property(name).CurrentValue;
+                }
+
+                return base[name];
+            }
+        }
+
+        public override object this[int i]
+        {
+            get
+            {
+                var name = _members[i];
+
+                return this[name];
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


borisdj:
I have looked into the problem and yours solution.
Have you done test with your fix? because now there is new problem, shadow Discriminator property is null and you can not insert it into DB since Discriminator column is not nullable.
`Column 'Discriminator' does not allow DBNull.Value.`

Regarding performance and overriding ObjectReader, maybe we should open an issue with FastMember project for Shadow property support.

Alexx999:
Well, you must add it to EF Core context you're trying to save it with to generate shadow values.
I was using EF generated keys so I had to add entities to context anyway, however, after a switch to DB generated keys I had the same issue.
Now I have following code in my save method:
```csharp
            if (entityType.Relational().DiscriminatorProperty != null)
            {
                ctx.ChangeTracker.AutoDetectChangesEnabled = false;
                ctx.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
                await ctx.AddRangeAsync(entities).ConfigureAwait(false);
            }
```

Alexx999:
Regarding FastMember - I was thinking about opening an issue for support of "extra" properties in ObjectReader (at least by making "current" field protected instead of private), however, FastMember isn't very active repo.

borisdj:
I have also seen that last change in FastMember repo was half year ago, but no harm opening issue, maybe author will have time in near future.

Also there is a way to do it without FastMember, instead of ObjectReader to use DataTable.
It's little more code, like 10 lines instead of 1 line: ObjectReader.Create(...
However this is not possible currently because .NetStandard 1.4 does not have DataTable, and I want library to tagret .NetStandard so it can be used with apps that are targeting different framework.
But I think .NetStandard 2.0, which will be released next month, will support it.
Still I wouldn't imediately make a switch because .NetStandard 2.0 will not work with Core 1.0, so it's better to wait some time before most developers update their apps.
.NetFramework verison support is the same, both .NetStandard 1.4 and 2.0 support .NetFramework 4.6.1+

borisdj:
Could you elaborate more on:
> Well, you must add it to EF Core context you're trying to save it with to generate shadow values.

And on the code snippet in that comment.
Where exactelly do you put it and when is are you calling your save method.
Because Bulk methods are communicating with DB without context's DbSets, and I have no Save method when doing Bulk ops.
So are you first disabling ChangeTracker, then add entities list to its DbSet and then call SaveChanges that you have overridden, and last call BulkInsert, or have I misunderstood it ?

Alexx999:
Yeah, you misunderstood a bit - I never call SaveChanges as it's enough to call AddRange() to get Discriminator set up (and IDs generated if any of them are being created by EF). I disable all tracking to improve performance - it's not really necessary.

borisdj:
You could now open an issue on FastMember repo regarding this.

paillave:
Hi @borisdj, any status for the support of inheritance without having to set the discriminator for each new instance?

borisdj:
There are 2 ways for it as explained in the README:
https://www.learnentityframeworkcore.com/model/shadow-properties

paillave:
Thanks for your answer @borisdj. Indeed, I missed this part, sorry for this. 
I still have a question for a bit more specific case that doesn't work out of the box:
Let suppose I have the following:
UmlDiagram
In my case I want to upsert cars using the businessKey as the reference (`BulkConfig.UpdateByProperties`). Giving the list of cars to `AddRange` sets automatic Ids to a negative number that is meant to be used by the `SaveChange` (I suppose!). The pb is that BulkExtensions tries to set this Id in the database... and of course, I get an exception.

borisdj:
Just set also PropertiesToExclude = new List<string> { nameof(Student.Id) }

paillave:
Yes I finally found out. Last pb: the retrieval of shadow properties makes the bulkload insanely slow: 30 seconds to bulkload 1000 rows against couple of milliseconds without shadow property.

borisdj:
Thx for the input. Will consider the options.

borisdj:
There is another way, if this issue is only because of _FastMember.ObjectReader_ it doesn't have to be used.
Alternative way with DataTable is already used for owned types:
https://github.com/borisdj/EFCore.BulkExtensions/blob/8fb0a4665f2fcbf5eef81460d1f6ad93a722e689/EFCore.BulkExtensions/SqlBulkOperation.cs#L40-L51
I could just add `bool UseDataTable` as optional setting in _BulkConfig_.

borisdj:
Another thing, this could be implemented even without adding new config element.
Just making a change it in a way that for Entites with Shadow properties DataTable is always used like it is for OwnedTypes.

borisdj:
This is now implemented with NuGet 2.3.9.
With Shadow properties DataTable is used insted of FastMember so there is no latency.

PR #22, opened 2017-11-13
Update SqlQueryBuilder.cs

HerrKater:
If the table was in a different schema (than dbo) it added an unnecessary "dbo." prefix, which was causing an error when the SetOutputIdentity was true.


borisdj:
I have merged your commit but made some changes, fix was moved to TableInfo:
`Schema = relationalData.Schema ?? "dbo";`
And then there is no need for explicit schema since it will be added with FULL property.
`q += $" OUTPUT INSERTED.* INTO {tableInfo.FullTempOutputTableName}";`

PR #26, opened 2018-01-09
Fix bulk inserting with option SetOutputIdentity

sperse-z:
**Steps to reproduce:**

1. Implement DB context with [global query filters](https://docs.microsoft.com/en-us/ef/core/querying/filters).

2. Execute bulk insert method with option SetOutputIdentity=true.

`Context.BulkInsert(entities, new BulkConfig { PreserveInsertOrder = true, BatchSize = 500, SetOutputIdentity = true});`

**Actual result:**
Method throws an exception.

**Expected result:**
No exception should be thrown.

### DIFF
--- a/EFCore.BulkExtensions/SqlQueryBuilder.cs
+++ b/EFCore.BulkExtensions/SqlQueryBuilder.cs
@@ -12,9 +12,9 @@ namespace EFCore.BulkExtensions
             return q;
         }
 
-        public static string SelectFromTable(string tableName, string orderByColumnName)
+        public static string SelectFromTable(string tableName)
         {
-            return $"SELECT * FROM {tableName} ORDER BY {orderByColumnName};";
+            return $"SELECT * FROM {tableName};";
         }
 
         public static string DropTable(string tableName)
--- a/EFCore.BulkExtensions/TableInfo.cs
+++ b/EFCore.BulkExtensions/TableInfo.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using System.Data;
 using System.Data.SqlClient;
 using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading.Tasks;
 using FastMember;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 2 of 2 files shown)


borisdj:
I am unable to reproduce your issue.

I have tried to test it by adding following line in `OnModelCreating` method:
`modelBuilder.Entity<ItemHistory>().HasQueryFilter(p => p.Remark != null);`
And test (`EFCoreBulkTest.RunInsert`) where `itemHistories` are Inserted with `SetOutputIdentity = true`
still passes.

Additionally I don't see how would filter on Query be related with _Select_, specifically Select with _OrderBy_.

Can you write entire StackTrace and Message of the Exception you get, and give more details about scenario that creates this problem.
Or if you could reproduce it in EFCore.BulkExtensions.Tests project like I have tried and post code snippet of it.

sperse-z:
You can't reproduce this bug because you configure global filter for entity 'ItemHistory'
`modelBuilder.Entity<ItemHistory>().HasQueryFilter(p => p.Remark != null);`

but call bulk insert method for entity 'Item'
`context.BulkInsert(entities, new BulkConfig { PreserveInsertOrder = true, SetOutputIdentity = true, BatchSize = 4000 }, (a) => WriteProgress(a));`

Try to specify a global filter for entity 'Item':
`modelBuilder.Entity<Item>().HasQueryFilter(p => p.Description != "1234");`

and you will receive an exception like this:
`The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.`

sql:
`SELECT [i].[ItemId], [i].[Description], [i].[Name], [i].[Price], [i].[Quantity], [i].[TimeUpdated]
FROM (
    SELECT * FROM dbo.[Item402d8032Output] ORDER BY ItemId;
) AS [i]
WHERE ([i].[Description] <> N'1234') OR [i].[Description] IS NULL`

borisdj:
I have managed to catch the same Exception.
And can confirm that your fix works.

I will merge it with base now.
Also _OrderBy_ can be done with _DynamicLinq_ (adding nuget _System.Linq.Dynamic.Core_) so I will consider changing your code, because with that there is no need for custom method.
All that is needed is to add `.OrderBy(this.PrimaryKeys[0])` in methods _UpdateOutputIdentity_ and its _Async_ version.
For example instead of:
`var entitiesWithOutputIdentity = QueryOutputTable<T>(context).ToList();`
we would have:
`var entitiesWithOutputIdentity = QueryOutputTable<T>(context).OrderBy(this.PrimaryKeys[0]).ToList();`

PR #30, opened 2018-01-18
Update SqlQueryBuilder.cs

HerrKater:
We have to check if the table exists in the tempDB before we drop it.


borisdj:
Just an advice for future, it's more practical to sent one pull request for connected code.
You can do multiple commits and then when all is done send a pull request.

HerrKater:
Thanks for accepting the pull request. Sorry about that, I just wanted to make a quick change, in the future I will create one pull request. Can you please publish the nuget package at some point, I really need this change.

borisdj:
You are welcome, and thank you for Contributing.
I will publish the nuget till end of this week.

PR #31, opened 2018-01-23
Add ability to keep identity columns as specified

vaindil:
`SqlBulkCopy` has the ability to do identity inserts, but the library didn't support this. I made these changes to the code to suit my needs, but I'm not sure if they will affect anything else. I also think this is a messy solution, I'd love any feedback you have.


borisdj:
From 2.0.6 version **KeepIdentity** is removed from **BulkConfig**.
BulkConfig now has **SqlBulkCopyOptions** which is Enum with [Flags] attribute that enables specifying one or more options:
_Default, KeepIdentity, CheckConstraints, TableLock, KeepNulls, FireTriggers, UseInternalTransaction_

PR #32, opened 2018-01-30
Added new logic to address a SQL collate issue.

obanion:
I identified an issue when trying to merge collated fields using bulk insert.  The current logic used the primary keys for the join condition which did not have the properly capitalize to location the correct column.  I changed the logic to use the column names of the primary keys instead of EF primary keys.


kluhman:
I also recently discovered this issue. I reported it in issue #33 

I also submitted PR #34 I think it addresses the same issue in your commit as well as more related issues further down the call tree. Would you mind taking a look at the PR and see if it resolves the issue you were seeing, then maybe we could close a PR to have just on fixing the issue. Thanks!

obanion:
I don't have a problem closing my PR if your fixes the problem I am having.    I left a few questions on your PR.

Issue #33, opened 2018-01-30
Throws exception on BulkInsert when SetOutputIdentity = true and the Primary Key db column name is different than the actual model property


kluhman:
I have a model defined like so 

```c#
public class Entity
{
    [Key]
    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
    public int Id { get; set; }

    //some other properties
}
```

However, using the fluent api in the context, the Id property is actually stored in the database as EntityId (name of the class + "Id").

When using SetOutputIdentity = true, it throws an exception saying "Invalid column: 'Id'", so it appears to be using the name from the model and not the actual column name somewhere.

kluhman:
PR submitted above for review.

borisdj:
Fix approved and PR merged.

PR #34, opened 2018-01-31
Fixes issue with BulkInsert when SetOutputIdentity = true and primary key column name is different than the property name

kluhman:
Fixes issue #33 

Was running a SQL Script checking for identity column and was using wrong column name throwing a SqlException. Uses the existing column mapping to map the primary key property to the actual column name

### DIFF
--- a/EFCore.BulkExtensions/SqlQueryBuilder.cs
+++ b/EFCore.BulkExtensions/SqlQueryBuilder.cs
@@ -31,7 +31,7 @@ namespace EFCore.BulkExtensions
         {
             string targetTable = tableInfo.FullTableName;
             string sourceTable = tableInfo.FullTempTableName;
-            List<string> primaryKeys = tableInfo.PrimaryKeys;
+            List<string> primaryKeys = tableInfo.PrimaryKeys.Select(k => tableInfo.PropertyColumnNamesDict[k]).ToList();
             List<string> columnsNames = tableInfo.PropertyColumnNamesDict.Values.ToList();
             List<string> nonIdentityColumnsNames = columnsNames.Where(a => !primaryKeys.Contains(a)).ToList();
             List<string> insertColumnsNames = tableInfo.HasIdentity ? nonIdentityColumnsNames : columnsNames;
--- a/EFCore.BulkExtensions/TableInfo.cs
+++ b/EFCore.BulkExtensions/TableInfo.cs
@@ -90,7 +90,7 @@ namespace EFCore.BulkExtensions
                     {
                         if (currentTransaction != null)
                             command.Transaction = currentTransaction.GetDbTransaction();
-                        command.CommandText = SqlQueryBuilder.SelectIsIdentity(FullTableName, PrimaryKeys[0]);
+                        command.CommandText = SqlQueryBuilder.SelectIsIdentity(FullTableName, PropertyColumnNamesDict[PrimaryKeys[0]]);
                         using (var reader = command.ExecuteReader())
                         {
                             if (reader.HasRows)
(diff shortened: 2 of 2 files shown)


EFCore.BulkExtensions/SqlQueryBuilder.cs
```diff
@@ -31,7 +31,7 @@ public static string MergeTable(TableInfo tableInfo, OperationType operationType
         {
             string targetTable = tableInfo.FullTableName;
             string sourceTable = tableInfo.FullTempTableName;
-            List<string> primaryKeys = tableInfo.PrimaryKeys;
+            List<string> primaryKeys = tableInfo.PrimaryKeys.Select(k => tableInfo.PropertyColumnNamesDict[k]).ToList();
```
obanion: I am not sure this will work.  The dictionary contains all the columns names however you are index the dictionary by primary keys.  If the column name is EntityId I am not sure you will find a result from tableInfo.PropertyColumnNamesDict["id"].

EFCore.BulkExtensions/TableInfo.cs
```diff
@@ -90,7 +90,7 @@ public void CheckHasIdentity(DbContext context)
                     {
                         if (currentTransaction != null)
                             command.Transaction = currentTransaction.GetDbTransaction();
-                        command.CommandText = SqlQueryBuilder.SelectIsIdentity(FullTableName, PrimaryKeys[0]);
+                        command.CommandText = SqlQueryBuilder.SelectIsIdentity(FullTableName, PropertyColumnNamesDict[PrimaryKeys[0]]);
```
obanion: Will this work with compounds keys?

EFCore.BulkExtensions/TableInfo.cs
```diff
@@ -130,7 +130,7 @@ public void CheckHasIdentity(DbContext context)
...
```
obanion: Will this work with compounds keys?

EFCore.BulkExtensions/SqlQueryBuilder.cs
kluhman: The PrimaryKeys collection contains a list of the primary key property names (as defined on the model), the PropertyColumnNamesDict maps all columns from the property name to the stored column name. For example, if we have the following model

```c#
public class Entity
{
    [Key]
    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
    [Column(Name = "EntityId")
    public int Id { get; set; }
}
```

Then:
PrimaryKeys == { "Id" }
PropertyColumnNamesDict == { new KeyValuePair<string, string>("Id", "EntityId") }

EFCore.BulkExtensions/TableInfo.cs
kluhman: It will not, however, compound keys are not currently supported by the library for this functionality. Also I don't think you can really have compound Identity columns, which is what this method is concerned with. You can have compound primary keys, but not identities.

EFCore.BulkExtensions/TableInfo.cs
kluhman: It will not, however, compound keys are not currently supported by the library for this functionality. Also I don't think you can really have compound Identity columns, which is what this method is concerned with. You can have compound primary keys, but not identities.

borisdj: This will work because as explained by @kluhman method CheckHasIdentity is called only when PK has Identity in which case PK can not compound from multiple columns.
So in this line you know for sure that PK is only one columns.

Issue #74, opened 2018-10-05
Add ability to know how many records were inserted and how many were updated on a BulkInsertOrUpdate call


fsmirne:
BulkInsertOrUpdate should have an option to return how many records were inserted and how many were updated. It seem like the code is already keeping track of this internally, but I believe there should be a way to extract that information after the call has been completed.
Thanks!

borisdj:
I assume you are doing insert/update by some unique column(s).
At the moment you can set _SetOutputIdentity_ to True in **BulkConfig** which will load back inserted data. This will give you Id created by Db (int type, autoincrement).
Now there is no explicit info which ones were inserted and which were updated, and I am not sure how to check for this in the library itself.
But I believe you can split that list into 2 these 2 group after BulkInsertOrUpdate().
One way would be to use Id(s). If it was _int_ type then before inserting get last Id and all those that have Id > LastId are inserted, others were update.
On the other hand If Id is of type _Guid_, SeqGuid created not in Db but in application, which you would have to set for all items in list, and then after function is executed those that have Id value, would be ones that were updated, because Id was loaded from existing record.

However I have another procedure for situation like this.
I first load records in memory into a dictionary where key is joined string of unique columns.
Then make separation of Insert group and Update group.
It is done so that Audit columns (CreatedTime, UpdatedTime) could be properly set.
This approach requires loading existing records into memory, options to do it:
1. Load all - feasible only if table is not very big.
2. Load only ones for update
    2.a) Use Join or Where with Contains (more info in ReadMe) - translated to Sql IN
    2.b) Use BulkRead - I am currently adding this method, and it will be published in a few days.

fsmirne:
Hey @borisdj thanks for your reply.
I was thinking more along the lines of [this solution](https://stackoverflow.com/questions/1268491/sql-server-2008-merge-best-way-to-get-counts).

Loading all records in memory is not a good solution for this case as the number of records can be very large. I think the most efficient way to get this info is to get the info as the operation is done as in the url above.

Thanks for your help!

borisdj:
I'll take a look.

fsmirne:
I cloned your repository and hacked the code a bit without any design in mind. I got it to work and performance was descent. I tested a few different scenarios using tempdb and not.
[...]

PR #77, opened 2018-10-09
Fix for Issue #74

fsmirne:


borisdj:
I have implemented this in a more simpler way.

PR #173, opened 2019-04-19
Update BatchUtil.cs

marekpow:
In some situations, the table may contain columns that you do not want to have in the model. For example, LastUpdate, RowVersion. Such columns can be filled by triggers or by the engine.

In this case, the GetSqlFragment method throws an exception (with the exception of a null reference) because it has no equivalent property to that column in the entity model.

To reproduce this exception, add RowVersion column to a table and try to update any column for the entity corresponding to this table by BatchUpdate method.

### DIFF
--- a/EFCore.BulkExtensions/BatchUtil.cs
+++ b/EFCore.BulkExtensions/BatchUtil.cs
@@ -85,23 +85,26 @@ namespace EFCore.BulkExtensions
                 var pArray = propertyName.Split(new char[] { '.' });
                 Type lastType = updateValuesType;
                 PropertyInfo property = lastType.GetProperty(pArray[0]);
-                object propertyUpdateValue = property.GetValue(updateValues);
-                object propertyDefaultValue = property.GetValue(defaultValues);
-                for (int i = 1; i < pArray.Length; i++)
+                if (property != null)
                 {
-                    lastType = property.PropertyType;
-                    property = lastType.GetProperty(pArray[i]);
-                    propertyUpdateValue = propertyUpdateValue != null ? property.GetValue(propertyUpdateValue) : propertyUpdateValue;
-                    var lastDefaultValues = lastType.Assembly.CreateInstance(lastType.FullName);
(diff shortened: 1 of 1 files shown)


borisdj:
I am not able to reproduce this.
In table Item I have added column: **TestCol nvarchar(MAX) Allow Nulls TRUE**.
And _BatchUpdate_ still works fine, this column remains with Null value.
Is that column not nullable maybe, and if it is how is regular insert then done?

marekpow:
Is this column added to the model? If yes then you can't reproduce this. You must add a column to the table but the app shouldn't know about its existence. In this case, the variable property in your code will have null value and application will throw an exception.

borisdj:
No it is not, I have added it only in Db, and the _BatchTest()_ passed.

marekpow:
Which way? This line of code returns null value in my app.

`PropertyInfo property = lastType.GetProperty(pArray[0]);`

How it can find the property, which doesn't exist in lastType? latType variable is the type of entity.

borisdj:
Maybe you have something specific in your app.
Try to reproduce it in test by running BatchTest().
Then execute following sql on Db:
`ALTER TABLE Item ADD TestCol nvarchar(MAX);`
Then run the same test again.

Regarding the code segment
```C#
PropertyInfo property = lastType.GetProperty(pArray[0]);
```
Here the _TestCol_ will not even appear in the list of Properties since it is not in entityType model at all.

marekpow:
Ok, I have few shadow properties. And this the cause of this error. Those properties are added to table context in OnCreating method, so they really exist in the table context but not in the entity class. So I think my pull request makes sense.

marekpow:
When the new version with my correction will be available?

marekpow:
Thank you :)

marekpow:
When you plan to publish a new version with my correction?

marekpow:
I ask you about the new nuGet package. The last is from 8 April.

borisdj:
You are welcome.
New NuGet will be in a day or two.

borisdj:
NuGet 2.4.6 published

Issue #177, opened 2019-04-23
Batch operations: Inner query parameters are passed in the clear


vladhorby:
When the original query used for the Batch operation contains parameters, they are passed in the clear (as values instead of query parameters) to the generated SQL query.
Example:
`var now = DateTimeOffset.UtcNow;`
`var query = context.Items.Where(x => x.Id <= maxBatchId);`
`await query.BatchUpdateAsync(new Item() { LastSeenOn = now });`

Expected query:
`UPDATE [item] SET [LastSeenOn] = @LastSeenOn
FROM [Items] AS [item]
WHERE [x].[Id] <= @__maxBatchId_0`

Actual query:
`UPDATE [item] SET [LastSeenOn] = @LastSeenOn
FROM [Items] AS [item]
WHERE [x].[Id] <= CAST(1000 AS bigint)`

This is a bigger problem when trying to optimize queries on Sql Server, because it doesn't recognize the queries as being the same.

I managed to find a solution, by modifying IQueryable.ToSql() to extract and pass the original parameters back to the Batch query generators.

`var parameterExpression = modelGenerator.ExtractParameters(diagnosticsLogger, query.Expression, parameterValues);`
`var queryModel = modelGenerator.ParseQuery(parameterExpression);`

I'll submit a PR with my changes.

borisdj:
Published NuGet 2.4.8

borisdj:
There is an issue with _ToParametrizedSql()_ when having Contains https://github.com/borisdj/EFCore.BulkExtensions/pull/203
I have made a temporary fix, in a way that ToParametrizedSql() in not used when having Contains (which maps to ' IN ' in sql):
https://github.com/borisdj/EFCore.BulkExtensions/blob/8e72267a5e1b0b08c7c3f977805dbf5e540aad31/EFCore.BulkExtensions/BatchUtil.cs#L66-L71
If you can take a look.

PR #178, opened 2019-04-23
Solves issue #177

vladhorby:
Batch operations: Inner query parameters are passed in the clear

When the original query used for the Batch operation contains parameters, they are passed in the clear (as values instead of query parameters) to the generated SQL query.
Example:
var now = DateTimeOffset.UtcNow;
var query = context.Items.Where(x => x.Id <= maxBatchId);
await query.BatchUpdateAsync(new Item() { LastSeenOn = now });

Expected query:
UPDATE [item] SET [LastSeenOn] = @LastSeenOn FROM [Items] AS [item] WHERE [x].[Id] <= @__maxBatchId_0

Actual query:
UPDATE [item] SET [LastSeenOn] = @LastSeenOn FROM [Items] AS [item] WHERE [x].[Id] <= CAST(1000 AS bigint)

Created an alternative to IQueryable.ToSql() - IQueryable.ToParametrizedSql() that builds the query using parameters instead of direct values. 
The SqlParameter list for the original query is returned as part of a Tuple together with the string Sql query. 
The batch operations use this list to populate their parameter lists before adding their own parameters.

Sorry for the mass whitespace changes - I realized too late that my Visual Studio reformatted the files.

### DIFF
--- a/EFCore.BulkExtensions/BatchUtil.cs
+++ b/EFCore.BulkExtensions/BatchUtil.cs
@@ -12,226 +12,226 @@ using Microsoft.EntityFrameworkCore.Query.Internal;
 
 namespace EFCore.BulkExtensions
 {
-	static class BatchUtil
-	{
-		// In comment are Examples of how SqlQuery is changed for Sql Batch
+    static class BatchUtil
+    {
+        // In comment are Examples of how SqlQuery is changed for Sql Batch
 
-		// SELECT [a].[Column1], [a].[Column2], .../r/n
-		// FROM [Table] AS [a]/r/n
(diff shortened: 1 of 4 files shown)


borisdj:
Some files (BatchUtil.cs and IQueryableExtensions.cs) differs significantly only because of different type of indentation.
You have used tab, and Project uses spaces, VisualStudio default.
So could you please cancel this PullRequest, change indentation type and make new P.R. so that real changes are easier visible.

vladhorby:
Sorry about that, I added a commit fixing the space / tabs issue. 
Only the real changes are visible now in the PR's "Files changed" tab.

PR #201, opened 2019-07-09
Async method enhancements

ysmoradi:
Added cancellation token support + Replaced non-async method calls with their async equivalent where possible + added ConfigureAwait(false) where missing.

### DIFF
--- a/EFCore.BulkExtensions.Tests/DbDataReaderExtensions.cs
+++ b/EFCore.BulkExtensions.Tests/DbDataReaderExtensions.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Data.Common;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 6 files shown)


borisdj:
I think it it more practical to add cancellation token in BulkConfig, so signature of many methods would Not have to be changes in all those places.
Can you make that change to this commit.

ysmoradi:
Good point, but some toolings such as ReSharper will advise developers to pass cancellation token if a method which is being called accepts that in its parameters directly.

borisdj:
We don't have to obey ReSharper advise.
BulkConfilg already has several params that are later forwarded to SqlBulkCopy.

ysmoradi:
Fair enough! ;D
Give me some time and I'll handle this.

borisdj:
I've decided to go with passing Cancellation token.
Commit merged.

ysmoradi:
Thanks in advance

PR #203, opened 2019-07-10
BatchDelete fails with Contains

halllo:
Hi. I really like this lib. Today I updated to the latest version and it seems to me there is a problem.
I need to perform a delete with an Contains check in the WHERE clause: 

`context.Items.Where(a => guidsToDelete.Contains(a.GuidId)).BatchDelete();`

This worked perfectly in Version 2.4.7 and started to fail in 2.4.8.
It seems to me that `query.ToParametrizedSql()` actually generates a wrong SQL statement:

image

I prepared a test to reproduce it but have no idea how to fix this. Do you have an idea?

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
@@ -17,6 +17,7 @@ namespace EFCore.BulkExtensions.Tests
             RunInsert();
             RunBatchUpdate();
             RunBatchDelete();
+            RunContainsBatchDelete();
 
             using (var context = new TestContext(ContextUtil.GetOptions()))
             {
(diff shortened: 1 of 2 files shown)


borisdj:
That part was recently added by another contributor.
@vladhorby  can you take a look.

borisdj:
Just initialize CProperty to `new C()` instead of leaving it _null_;
You can do it after setting _CarCount_ or in a separate loop before calling BulkUpdate:

```C#
foreach (var entity in list)
{
    entity.CProperty = new C();
}
```

PS Upgrade NuGet to latest version since there was another issue with this that is now fixed.

borisdj:
With version 2.5.0 I have made a temporary fix, in a way that _ToParametrizedSql()_ in not used when having Contains, which maps to ' IN ' in sql.

PR #223, opened 2019-09-14
Implemented common progress setting logic

Janek91:
There was no progress reporting for SQLite, so I provided one and made it common with SQL Server one.


borisdj:
Thx for contrib.

PR #232, opened 2019-09-24
BatchDelete fails with Contains and empty list

halllo:
Hi. I really like this lib. Yesterday I updated to the latest version (2.6.3) and it seems to me there is a problem. I need to perform a delete with a Contains check in the WHERE clause:

`context.Items.Where(a => guidsToDelete.Contains(a.GuidId)).BatchDelete();`

This worked perfectly in Version 2.4.7 and started to fail in 2.4.8 (https://github.com/borisdj/EFCore.BulkExtensions/pull/203) which is why I didn't update until yesterday.

It seemed to work but in case of `guidsToDelete` being an empty list it fails again:

> System.ArgumentException : No mapping exists from object type System.Collections.Generic.List`1[[System.Guid, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]] to a known managed provider native type.

I don't know how to fix this, so this PR offers just a test to reproduce. Do you have an idea?

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
@@ -22,6 +22,8 @@ namespace EFCore.BulkExtensions.Tests
             RunBatchUpdate();
             RunBatchDelete();
             RunContainsBatchDelete();
+            RunContainsBatchDelete2();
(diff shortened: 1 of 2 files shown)


halllo:
OK, I will do it.

But this was not necessary in v2.4.7. Many people will not think about it. This is why I consider this a bug.

PR #258, opened 2019-11-13
Feature/238 batch operations with top sql statement

sky-daniel-paula:
Support "TOP(xx)" statement for SQL Server in batch delete and batch update.
This is useful for scenarios in which we have for example a recurring job to delete just a few records each time it runs.

Really like this extension, great job 👍


borisdj:
Pull merged and NuGet 3.0.3 published.
Thx for contrib.

PR #259, opened 2019-11-14
[238] add support to TOP statement for SQL server in EFCore 2.2

sky-daniel-paula:
Support "TOP(xx)" statement for SQL Server in batch delete and batch update in EFCore 2.2.
This is useful for scenarios in which we have for example a recurring job to delete just a few records each time it runs.

Really like this extension, great job 👍


osipenkom:
@borisdj Hello, I'm using your extension and it is very usefull for me. Is it possible to merge this functionality to 2.4.3 version and deploye it to Nuget as 2.4.3.1 for example, so I and my teammates will be able to use this DELETE TOP X functionality which is very needed for us?

borisdj:
2.6.4 published

PR #279, opened 2019-12-29
support for filling columns from derived entities

NQuirmbach:
This allows to insert/update from derived entity colums even an enumerable from the abstract base type is passed into the function


borisdj:
Thx for contrib.

PR #284, opened 2020-01-12
Support IEnumerable<>.Contains and Any query.

huan086:
Solution is inspired by

https://github.com/dotnet/efcore/blob/master/src/EFCore.Relational/Query/Internal/QueryingEnumerable.cs#L107

https://github.com/dotnet/efcore/blob/master/src/EFCore.Relational/Query/Internal/RelationalCommandCache.cs#L44


huan086:
I can't reach the case where `relationalCommandCache == null`, thus I have maintained the existing behavior.

borisdj:
Thx for contribution.

borisdj:
NuGet 3.1.1 published.

PR #296, opened 2020-02-20
Support long for identity type in sqlite

bjaminp:
There is an error thrown right now when you have an identity column that is a long type because the identity is explicitly cast to an int. I changed it to only cast to an int if the identity is an int.


borisdj:
Thx for contrib.

PR #306, opened 2020-03-12
Runtime entity type support

konzen:


borisdj:
Could you elaborate more on the made changes and when do you need Runtime entity,
what's the use case for the support you added?

konzen:
I'm using the System.Linq.Dynamic.Core package to create dynamic queries from texts, so there's no way to know the type of the entity at design time and using EFCore.BulkExtensions to do mass CRUD operations.

The logic change was small. Override versions were created for entities in which the type is unknown at design time.

IQueryable<T> was replaced for IQueryable in sources when the type of the entity is not relevant or can be obtained in another way.

When the type of the entity cannot be neglected, I created an overrite version of the function replacing IQueryable<T> queryable with IQueryable queryable, Type entityType and in the case of IList<T> entities with Type entityType, IList<object> entities.

borisdj:
Could you also add a Tests with examples of all the scenarios mentioned above.

borisdj:
Merged
Thx for contrib

borisdj:
Can you add example or one Test with this use case, dinamic queries without defined Entity.
I want to add note about it in ReadMe but not sure how to explain its usage.

borisdj:
@konzen I'am doing code refactoring and want to simplify some procedures.
In that regard I need to join methods into a single flow, both generic<T> and Runtime that your PR has added.
But I can find a way to use runtime methods from test, how it that called with dynamic ?

borisdj:
Generic T and object methods are now merged into single methods with optional arg: **type**.
v3.5.2+

Issue #281, opened 2020-01-08
Calling BulkInsertOrUpdateAsync() will hang when running in Linux


robolize:
I have been trying to get my data process to work when deploying to a Linux container in Cloud Foundry.  I am using the "dotnet_core_buildpack" buildpack and "cflinuxfs3" stack.  It's running DotNetCore 3.1 and targeting "linux-x64" self contained.  I have tried many things to get this to work, but the process would always hang when calling BulkInsertOrUpdateAsync().  Sorry, I don't know how to provide any specific details.  I'm not getting an exception or anything.  All I have to go on is logging and that stops giving any messages when it hangs.  Please let me know if there is a way to debug this further.
I was able to get it to work by changing over to the "binary_buildpack" buildpack and targeting "win10-x64" self contained.
Again, sorry I don't know how to provide more details, but I at least wanted to report the issue.

Maleaume:
I have same issue. It's take a lot of time under Linux:
1,5 M items to insert on Windows:  few minutes
1.5 M items to insert on Linux :  one night is not enough ....:(

Numpsy:
Could this issue be down to https://github.com/dotnet/SqlClient/issues/262 in the MS Sql Client?

Might be worth updating the Microsoft.Data.SqlClient reference from 1.10 to 1.1.1 to get the fix for that?

Maleaume:
Thanks  @Numpsy 
 I have updated nuget dependencies of EF.BulkExtensions to 1.1.1. It's working fine now
Maybe it will be interesting to update EFCore.BulkExtensions packages (@borisdj )

libe:
Hello,

this is very reproducible if you do a bulk insert with more 200MB of data with two Task in parallel. It hangs with a deadlock under Linux keeping the TCP connection open without sending any packets.

We got rid of this issue by updating the SQL Client to version 1.1.1.

borisdj:
Merged and NuGet 3.1.1 published.

Maleaume:
Thanks a lot :)

PR #309, opened 2020-03-17
Update Microsoft.Data.SqlClient reference to version 1.1.1

Numpsy:
As described at https://github.com/dotnet/SqlClient/issues/262, version 1.1.0 of Microsoft.Data.SqlClient has a bug that can cause deadlocks in async calls, which is fixed in the 1.1.1 release.

refs #281.


borisdj:
Already updated with previous PullRequest so can't merge the same version, still Thx for contrib.

Numpsy:
Hi,

Not sure what you mean - the project in the master branch still looks to refer to 1.1.0?

borisdj:
True, maybe it was another NuGet.
I am doing more then one update.

Issue #253, opened 2019-10-31
Bulk-Update closes an open SqlConnection that leads to "sp_reset_connection"


PawelGerr:
`MergeAsync` or rather `CheckHasIdentityAsync` that is used by `BulkUpdateAsync` is closing the `SqlConnection` previously opened by some other code. This behavior leads to (multiple) executions of `exec sp_reset_connection` that drops temp tables in the middle of a (ASP.NET) request. 

An option could be: Instead of using the `SqlConnection` directly you may use `OpenConnectionAsync` on `context.Database`. Doing that you don't need to check the connection state because [RelationalConnection](https://github.com/aspnet/EntityFrameworkCore/blob/f780a3aef9839131aef118cd73dda625e22fb980/src/EFCore.Relational/Storage/RelationalConnection.cs) is using a counter `_openedCount` internally. So, an open connection stays open and a closed connection will be opened and closed in the end.

```
public async Task CheckHasIdentityAsync( .... )
{
   await context.Database.OpenConnectionAsync().ConfigureAwait(false);

   try
   {
     // content of "CheckHasIdentityAsync" without "sqlConnection.OpenAsync(cancellationToken)"
   }
   finally
  {
     context.Database.CloseConnection();
  }
}
```

I'm using 
* EFCore.BulkExtensions: 2.6.1
* Microsoft.EntityFrameworkCore.SqlServer: 2.2.6

borisdj:
Can you try now with latest version.
If the issue repeats write a test where it is reproducible.

videokojot:
Hi, I am having issues with transactions being completed by BulkOperations, while it should not be either commited or rolledback, but I cannot understand why it is happening, maybe it is connected to this issue. 

@PawelGerr can I ask you how is the CheckHasIdentityAsync closing the previously opened connection?

PawelGerr:
@videokojot There is a pull request for that https://github.com/borisdj/EFCore.BulkExtensions/pull/314
There are multiple places in the code that close the connection, it is not just CheckHasIdentityAsync.
If you want to be sure that your issue has something to do with mine then you can use my fork to check it out: https://github.com/PawelGerr/EFCore.BulkExtensions

videokojot:
Thanks, I see what could cause it. I will check it out. But I am calling BulkOperatiosn via wrapper, which opens the transaction everytime, so your fix should not affect mine problem, because it seems, he is not closing connection if there if transaction present. I ll keep diggin.

[...]

PR #314, opened 2020-03-30
Bulk-Operations should not close an open connection

PawelGerr:
PR for issue #253 

The issue still exists in the new version. I've provided tests and bugfixes for both the sync and async operations. There were multiple occurrences that closes the connection that was opened by "someone else" before but that may be not all of them.

Please backport the fixes to version 2.x if the solution is acceptable for you because a lot of companies are still using EF Core 2.x. Thx!

btw. there are/were 4 tests that are failing before my change.

PS: I've added an .editorconfig so the code style of the contributors are maching yours.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+indent_style = space
+
+[*.cs]
+indent_size = 4
+tab_width = 4
+end_of_line = lf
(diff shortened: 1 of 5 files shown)


borisdj:

PawelGerr:
I've rebased my changes.

borisdj:
Thx for Contrib.

Issue #295, opened 2020-02-20
BulkInsert fails with SetOutputIdentity on Sqlite with long type


bjaminp:
EF Core 3.1.2
EFCore.BulkExtensions 3.1.0

I am using sqlite with EFCore.

When using the following bulk insert I get an error.
```
var bulkConfig = new BulkConfig {PreserveInsertOrder = true, SetOutputIdentity = true};
this.BulkInsert(parts, bulkConfig);
```
My identity field is a long type.

Error:
`Unable to cast object of type 'System.Int32' to type 'System.Int64'.`
```
   at FastMember_dynamic.Part_1.set_Item(Object , String , Object )
   at EFCore.BulkExtensions.SqlBulkOperation.Merge[T](DbContext context, IList`1 entities, TableInfo tableInfo, OperationType operationType, Action`1 progress)
   at EFCore.BulkExtensions.DbContextBulkTransaction.Execute[T](DbContext context, IList`1 entities, OperationType operationType, BulkConfig bulkConfig, Action`1 progress)
   at EFCore.BulkExtensions.DbContextBulkExtensions.BulkInsert[T](DbContext context, IList`1 entities, BulkConfig bulkConfig, Action`1 progress)
```

borisdj:
Fixed with 3.1.1

Numpsy:
I think I just fell over this issue in 3.1.1 when using async bulk inserts.
Looking at the code, it looks like it might have been fixed in the Merge function, but not the MergeAsync function?

JasonLandbridge:
This is unfortunately still an issue. If I have the following

```
var bulkConfig = new BulkConfig {PreserveInsertOrder = true, SetOutputIdentity = true};
await this.BulkInsertAsync(parts, bulkConfig, cancellationToken);
```
I get the following exception
```
System.InvalidCastException: Unable to cast object of type 'System.Int64' to type 'System.Int32'.
   at lambda_method(Closure , Object , Object )
   at EFCore.BulkExtensions.FastProperty.Set(Object instance, Object value)
   at EFCore.BulkExtensions.SqlBulkOperation.MergeAsync[T](DbContext context, Type type, IList`1 entities, TableInfo tableInfo, OperationType operationType, Action`1 progress, CancellationToken cancellationToken)
[...]

PR #325, opened 2020-04-21
Extend the fix for #295 to SqlBulkOperation.MergeAsync

Numpsy:
I was doing some tests with BulkInsertAsync using Sqlite and fell across what looks like #295, which I think might be because #296 only fixed the sync version of Merge.

This PR just applies the same change to MergeAsync. (which seems to fix my tests, though I don't know enough about how it works to say if this is the best approach)


borisdj:

Numpsy:
updated

borisdj:
Thx for Contrib.

PR #331, opened 2020-05-02
Provide helpful exception message when table contains no primary key

stevekirks:
Currently when an entity type with no Key is used it would throw a NullReferenceException. This change instead throws an InvalidOperationException with message "EntitySet for Type: { type.Name } must contain a Primary Key"


borisdj:

stevekirks:
@borisdj , I've resolved the current conflict with the main branch now. Thanks

borisdj:
Thx for Contrib.

PR #336, opened 2020-05-06
Update SqlBulkOperation.cs for SQLite

davidstjacques:
When using SQLite and a custom Converter that ouputs a null value, we must also convert it to DBNull.Value or it crashes.


borisdj:
Thx for Contrib.

PR #355, opened 2020-07-03
Fix NullReferenceException on bulk operations on entities with no keys

gandhis1:
@borisdj 

I tested this with an `InsertAsync` operation on an entity that was configured with `HasNoKey()`. It appears to work successfully. That said, I don't know the full context of this code, but it seems in line with the recent changes made in 121d2d5.

### DIFF
--- a/EFCore.BulkExtensions/SqlBulkOperation.cs
+++ b/EFCore.BulkExtensions/SqlBulkOperation.cs
@@ -121,6 +121,7 @@ namespace EFCore.BulkExtensions
                     {
                         var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
 
+                        type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
                         var typeAccessor = TypeAccessor.Create(type, true);
                         int rowsCopied = 0;
                         foreach (var item in entities)
@@ -228,6 +229,7 @@ namespace EFCore.BulkExtensions
                     {
                         var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
 
+                        type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
                         var typeAccessor = TypeAccessor.Create(type, true);
                         int rowsCopied = 0;
 
@@ -342,6 +344,7 @@ namespace EFCore.BulkExtensions
                     {
                         var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
 
+                        type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
                         var typeAccessor = TypeAccessor.Create(type, true);
                         int rowsCopied = 0;
                         foreach (var item in entities)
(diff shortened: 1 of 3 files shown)


gandhis1:
@knalinne I also included a fix for entities which do not have an identity column going through an irrelevant branch in `SqlQueryBuilderSqlite.cs`.

EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -483,6 +486,7 @@ public static async Task MergeAsync(DbContext context, Type type, IList<object>
                     {
                         var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
 
+                        type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
```
gandhis1, line 489: The `type` that is passed in is always the abstract base class if the entity being passed in is covariant. So for instance `BulkInsert<T>` at compile time recognizes T as `AbstractBaseClass`, but at run time you may be passing in a type of `DerivedClass`.

EFCore.BulkExtensions/SqlQueryBuilderSqlite.cs
```diff
@@ -17,7 +17,7 @@ public static string InsertIntoTable(TableInfo tableInfo, OperationType operatio
             List<string> columnsList = tableInfo.PropertyColumnNamesDict.Values.ToList();
 
             bool keepIdentity = tableInfo.BulkConfig.SqlBulkCopyOptions.HasFlag(SqlBulkCopyOptions.KeepIdentity);
-            if(operationType == OperationType.Insert && !keepIdentity)
+            if(operationType == OperationType.Insert && !keepIdentity && tableInfo.HasIdentity)
```
gandhis1, line 20: In the absence of this check, on line 22, `tableInfo.IdentityColumnName` will be null, leading to a `NullReferenceException` when indexing `tableInfo.PropertyColumnNamesDict`.

EFCore.BulkExtensions/TableInfo.cs
```diff
@@ -123,7 +123,7 @@ private void LoadData<T>(DbContext context, Type type, IList<T> entities, bool l
             bool AreSpecifiedUpdateByProperties = BulkConfig.UpdateByProperties?.Count() > 0;
             var primaryKeys = entityType.FindPrimaryKey()?.Properties?.Select(a => a.Name)?.ToList();
 
-            HasSinglePrimaryKey = primaryKeys.Count == 1;
+            HasSinglePrimaryKey = primaryKeys?.Count == 1;
```
gandhis1, line 126: The previous updates fixed line 124 with conditional null access, but this line, which is also susceptible to a potential null value on `HasNoKey()` entities, was omitted. This is fixed here.

EFCore.BulkExtensions/TableInfo.cs
```diff
@@ -132,9 +132,9 @@ private void LoadData<T>(DbContext context, Type type, IList<T> entities, bool l
             if (entityType.IsAbstract())
             {
                 var extendedAllProperties = allProperties.ToList();
-                foreach (var dervied in entityType.GetDirectlyDerivedTypes())
+                foreach (var derived in entityType.GetDirectlyDerivedTypes())
                 {
-                    extendedAllProperties.AddRange(dervied.GetProperties());
+                    extendedAllProperties.AddRange(derived.GetProperties());
```
gandhis1, line 137: Merely fixing a typo.

borisdj:
Merged and 3.1.4 published

Issue #356, opened 2020-07-06
SQL Sever implementations use the existing transaction while Sqlite does not


gandhis1:
@borisdj 

In all of the SQL Server operations, the ambient transaction will be used. For example:

https://github.com/borisdj/EFCore.BulkExtensions/blob/1d24af75300eb053a19cf4c2487172d94dc7cf09/EFCore.BulkExtensions/SqlBulkOperation.cs#L60-L66

However in the Sqlite branches, if a `SqliteTransaction` is not specified, a new one is created:

https://github.com/borisdj/EFCore.BulkExtensions/blob/1d24af75300eb053a19cf4c2487172d94dc7cf09/EFCore.BulkExtensions/SqlBulkOperation.cs#L112-L119

1) Why doesn't it use the existing transaction? Making use of the extension method, this would be the following line of code:

`var transaction = (SqliteTransaction)dbContext.Database.CurrentTransaction.GetUnderlyingTransaction()`

2) If, per 1, it can sniff out the ambient `SqliteTransaction`, under what circumstances are the config parameters needed?

3) Sqlite notoriously cannot handle nested transactions, so if you are mocking a SQL Server connection with a Sqlite in-memory database, which I believe will be a very common application, and you wrap your bulk operations in a DbContext transaction, then you will hit this issue and will have to make your code aware of the Db provider. Which totally undermines the whole point of entity framework, in that the db context is dependency injected so your code does not need to be intelligent / coupled.

Generally speaking, I'm trying to understand why the special Sqlite configuration parameters are needed. I would ideally like to be able to switch seamlessly between SQL Server and Sqlite without any code changes needed.

borisdj:
1. Initially I wasn't able to make it work properly with internal transaction, probably did wrong cast or something like that.
So it was exposed via config.
I see you have manage to integrate it successfully, so I have merged your PullRequest.
Will see now to clean the code and remove it from BulkConfig.
Thx for contrib.

PR #358, opened 2020-07-08
Use ambient SqliteTransaction if available

gandhis1:
Fixes https://github.com/borisdj/EFCore.BulkExtensions/issues/356. Tagging everyone who has touched this code before @borisdj @PawelGerr 

T**his is a fundamental change so it definitely requires some review/discussion prior to merge**. The diff looks like a lot but this is merely indentation changes - I will annotate the exact lines which have actually changed.

In summary - SQL Server supports the following code out of the box, and as far as I can tell, does NOT use a nested transaction below the application's transaction below:

```
using var transaction = dbContext.Database.BeginTransactionAsync();
dbContext.BulkInsert(entities);
transaction.Commit();
```

*This code does NOT work out of the box with Sqlite.* This is because you need to explicitly specify the `SqliteTransaction` (conversely `SqliteConnection` is unnecessary). I cannot fathom why this should be required, but the reason it is, is because the Sqlite implementation chooses to create a new transaction and then commit it, and *Sqlite does not support nested transactions*.

The code here removes the nested transaction, and always uses the ambient transaction. I would even argue as far as deprecating the `SqliteConnection` and `SqliteTransaction` parameters. Unless I'm missing something, I just don't understand what the need is for these when using Sqlite and why this need does not exist when using SQL Server.

Perhaps I lack a full understanding of the rationale behind these features, as all I can see from my vantage point is two implementations that are inconsistent, and therefore require that application code be "smart" and aware of the db context provider. The whole point of Entity Framework is that application code should be decoupled from the provider, so I think it would be optimal to be able to switch out Sqlite and SQL Server with no change in application code.

### DIFF
--- a/EFCore.BulkExtensions/SqlBulkOperation.cs
+++ b/EFCore.BulkExtensions/SqlBulkOperation.cs
@@ -116,28 +116,18 @@ namespace EFCore.BulkExtensions
 
                 try
                 {
-                    var transaction = tableInfo.BulkConfig.SqliteTransaction ?? connection.BeginTransaction();
-                    try
-                    {
-                        var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
+                    var transaction = tableInfo.BulkConfig.SqliteTransaction ?? (SqliteTransaction)context.Database.CurrentTransaction.GetUnderlyingTransaction(tableInfo.BulkConfig);
 
-                        type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
-                        var typeAccessor = TypeAccessor.Create(type, true);
-                        int rowsCopied = 0;
-                        foreach (var item in entities)
-                        {
-                            LoadSqliteValues(tableInfo, typeAccessor, item, command);
-                            command.ExecuteNonQuery();
-                            SetProgress(ref rowsCopied, entities.Count, tableInfo.BulkConfig, progress);
-                        }
-                    }
-                    finally
+                    var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
+
+                    type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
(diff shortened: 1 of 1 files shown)


EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -116,28 +116,18 @@ private static void Insert<T>(DbContext context, Type type, IList<T> entities, T
 
                 try
                 {
-                    var transaction = tableInfo.BulkConfig.SqliteTransaction ?? connection.BeginTransaction();
-                    try
-                    {
-                        var command = GetSqliteCommand(context, type, entities, tableInfo, connection, transaction);
+                    var transaction = tableInfo.BulkConfig.SqliteTransaction ?? (SqliteTransaction)context.Database.CurrentTransaction.GetUnderlyingTransaction(tableInfo.BulkConfig);
```
gandhis1, line 119: In each of the 4 Sqlite operations, this line has changed to use the ambient transaction.

EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -116,28 +116,18 @@ private static void Insert<T>(DbContext context, Type type, IList<T> entities, T
...
+                    type = tableInfo.HasAbstractList ? entities[0].GetType() : type;
+                    var typeAccessor = TypeAccessor.Create(type, true);
+                    int rowsCopied = 0;
+                    foreach (var item in entities)
                     {
-                        if (tableInfo.BulkConfig.SqliteTransaction == null)
-                        {
-                            transaction.Commit();
-                            transaction.Dispose();
-                        }
```
gandhis1, line 140: And the commit code, along with the entire `try`-`finally` block, was removed. Nothing else changed, GitHub is just showing indentation changes.

PR #370, opened 2020-08-11
Various small fixes including warnings, code styles and spellings

TheCactusBlue:


EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -866,28 +866,25 @@ internal static SqliteCommand GetSqliteCommand<T>(DbContext context, Type type,
...
+                    break;
+                case OperationType.InsertOrUpdateDelete:
+                    throw new NotSupportedException("Sqlite supports only UPSERT(analog for MERGE WHEN MATCHED) but does not have functionality to do: 'WHEN NOT MATCHED BY SOURCE THEN DELETE'" +
+                                                    "What can be done is to read all Data, find rows that are not is input List, then with those do the BulkDelete.");
+                case OperationType.Update:
+                    command.CommandText = SqlQueryBuilderSqlite.UpdateSetTable(tableInfo);
+                    break;
+                case OperationType.Delete:
+                    command.CommandText = SqlQueryBuilderSqlite.DeleteFromTable(tableInfo);
+                    break;
```
gandhis1, line 887: Can you use the C# 8.0 switch expression syntax here? I think it would be even cleaner. The exception state could be done in an `if` statement prior to the switch expression

EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -19,7 +19,7 @@ public enum DbServer
     {
         SqlServer,
         MySql,
-        PostrgeSql,
+        Npgsql,
```
gandhis1, line 22: Not sure if a .NET specific library is the best name here - everything else is the generic name of the database. That said it was mis-spelled previously, should have "PostgreSQL"

Is this enumeration element even used yet, anyway?

EFCore.BulkExtensions/BatchUtil.cs
```diff
@@ -16,9 +16,19 @@
...
+    // public class BatchSql
+    // {
+    //     public string Sql;
+    //     public string TableAlias;
+    //     public string TableAliasSufixAs;
+    //     public string TopStatement;
+    //     public string LeadingComments;
+    //     public IEnumerable<object> InnerParameters;
+    // }
+
```
gandhis1, line 28: Is this intended to be commented out? Why is it here?

TheCactusBlue: wait, I forgot to remove that from my first commit: I was originally trying to write a PostgreSQL extension for EFCore BulkExtensions. Yeah, that can be removed.

EFCore.BulkExtensions/SqlBulkOperation.cs
TheCactusBlue: Currently working on another PR to add PostgreSQL support right now, although it'll take more time than this one.
Anyways, reason for the name "Npgsql" is to keep it consistent with other DbServer string, where `(providerName.EndsWith(DbServer.SqlServer.ToString())` is used by SQL server provider, although I can fix that as well if required.

EFCore.BulkExtensions/SqlBulkOperation.cs
TheCactusBlue: Yeah, I can change that.

borisdj: C# 8.0 can be used.

EFCore.BulkExtensions/SqlBulkOperation.cs
borisdj: That Enum is DbServer so it should stay as it as, with the correction Post**rg**eSql -> Post**gr**eSql.
And ProviderName can be added as an optional Attribute of this Enum so it can be defined when not the same, like Npgsql for PostgreSQL.
Thx for contrib.

borisdj:
Btw there is Diff with entire file EFCoreBatchTest.cs even though it is the same, maybe because of indentation or something like that.
Could you sort that out so that only real changes are in Diff.

borisdj:
Syntax changes integrated, thx for suggestions.

Issue #412, opened 2020-10-15
Exception "Value cannot be null. (Parameter 'method')" - BulkUpdateAsync with baseclass private properties


jr01:
The fix that was done in https://github.com/borisdj/EFCore.BulkExtensions/issues/393 doesn't take properties that exist on a baseclass into account.  

```c#
public abstract class CompanyEntity
{
     public System.Guid CompanyId { get; private set; }

     public Company Company { get; private set; }
}

public class SomeEntity : CompanyEntity
{
	public boolean SomeProperty { get; set; }
}

... dbContext.SomeEntities
                .BatchUpdate(
                    x => new SomeEntity
                    {
                        SomeProperty = true
                    }
```

```
    System.ArgumentNullException: Value cannot be null. (Parameter 'method')
  Stack Trace: 
    Expression.Call(Expression instance, MethodInfo method, Expression arg0)
    Expression.Call(Expression instance, MethodInfo method, IEnumerable`1 arguments)
    Expression.Call(Expression instance, MethodInfo method, Expression[] arguments)
    FastProperty.InitializeSet()
    TableInfo.LoadData[T](DbContext context, Type type, IList`1 entities, Boolean loadOnlyPKColumn)
    TableInfo.CreateInstance[T](DbContext context, Type type, IList`1 entities, OperationType operationType, BulkConfig bulkConfig)
    TableInfo.CreateInstance(DbContext context, Type type, IList`1 entities, OperationType operationType, BulkConfig bulkConfig)
    BatchUtil.GetSqlUpdate[T](IQueryable query, DbContext context, Type type, Expression`1 expression)
    BatchUtil.GetSqlUpdate[T](IQueryable`1 query, DbContext context, Expression`1 expression)
    IQueryableBatchExtensions.BatchUpdateAsync[T](IQueryable`1 query, Expression`1 updateExpression, CancellationToken cancellationToken)
```

See also https://stackoverflow.com/questions/31807874/cant-get-getsetmethod-for-property-with-private-setter-from-base-class

borisdj:
Should be fixed now with 3.2.5

jr01:
Excellent. Thanks!

PR #413, opened 2020-10-15
Bugfix/private baseclass getter setter

jr01:
Should fix https://github.com/borisdj/EFCore.BulkExtensions/issues/412


borisdj:
I have already fixed this, according to the link you provider in the issue.
Try now the latest version 3.2.5
Thx for contrib.

Issue #88, opened 2018-10-28
Geometry , Geography 


Olbrasoft:
https://blogs.msdn.microsoft.com/dotnet/2018/10/17/announcing-entity-framework-core-2-2-preview-3/

Example solution:
https://github.com/Olbrasoft/TravelCore/blob/master/src/Olbrasoft.Data.Entity.Framework.Bulk/ObjectReaderEx.cs

This solution is not optimal Nuget  System.Data.SqlTypes  for this code:
                  case NetTopologySuite.Geometries.Point dbPoint:
                    {
                        var chars = new SqlChars(dbPoint.AsText());
                        return SqlGeography.STGeomFromText(chars, dbPoint.SRID);
                    }
                case NetTopologySuite.Geometries.Polygon dbPolygon:
                    {
                        var chars = new SqlChars(dbPolygon.AsText());
                        return SqlGeography.STGeomFromText(chars, dbPolygon.SRID);
                    }
Throw warnning message not compatible net Core , but it works.

borisdj:
What is the issue here exactly, did you have some exception with these Geo. types?

Olbrasoft:
https://www.nuget.org/packages/Microsoft.SqlServer.Types
Only in Visual Studio is message :
This package may not be fully compatible with your project

comment:
In Entity Framework 6  DbGeography  => Translatet to SqlGeography
In Entity Framework Core https://github.com/NetTopologySuite/NetTopologySuite
I did not have time to examine it anymore, but I just cut the import and I hope you will solve the struggle with time :)

borisdj:
When I catch enough time I'll take a look.

acordmike:
FYI, the EntityFramework.BulkInsert project has this functionality, so you  may be able to look at the code for some guidance.
On specific file is https://github.com/ghost1face/EntityFramework.BulkInsert/blob/master/src/EntityFramework.BulkInsert/Helpers/MappedDataReader.cs

4nonym0us:
Would be great to have support for Geometry\Geography sql types. Is it still on the roadmap?
The issue became more relevant since EF Core supports working wit spatial data and is compatible with NetTopologySuite.

lettucebo:
Have this functionality would be very helpful~

hieuxlu:
[...]

PR #420, opened 2020-10-26
Implement handling of Geometry type

Farami:
This allows bulk inserting and updating of NetTopologySuites Geometry type.
By default it will use SRID 4326, which is the default SRID for MSSQL.

### DIFF
--- a/EFCore.BulkExtensions/SqlBulkOperation.cs
+++ b/EFCore.BulkExtensions/SqlBulkOperation.cs
@@ -11,6 +11,8 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Threading;
 using System.Threading.Tasks;
+using NetTopologySuite.Geometries;
+using NetTopologySuite.IO;
 
 namespace EFCore.BulkExtensions
 {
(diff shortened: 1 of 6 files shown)


Farami:
This will resolve #88.

tejsj:
+1

This feature would be so nice!

EFCore.BulkExtensions/SqlBulkOperation.cs
```diff
@@ -736,6 +744,11 @@ internal static DataTable GetDataTable<T>(DbContext context, Type type, IList<T>
                         propertyType = underlyingType;
                     }
 
+                    if (propertyType == typeof(Geometry) && isSqlServer) 
+                    {
```
valdas1, line 748: Add a check for propertyType equals geometry, and change sqlServerBytesWriter (line 718) IsGeography property to false.
```
if propertyType == typeof(Geometry)
{
    sqlServerBytesWriter.IsGeography = false;
}
```

valdas1 (changes requested):
Changes to EFCore.BulkExtensions/SqlBulkOperation.cs

I have a question about setting the IsGeography property of SqlServerBytesWriter to true, and the handling of the differences between geometry and geography types.

I'm not quite sure the current version would work, so if anyone could do a quick review, or write a comment, that would be great.

borisdj:
I have merged some previous PullRequest, so could you first resolve current conflicts.
Thx for contrib.

miloszskalecki:
Hi Guys, any news on this fix?

borisdj:
I have managed to manually integrate the proposed solution to the latest version.

PR #435, opened 2020-12-10
Fix bug in batch update and batch delete with User Defined Table Type parameter

daujyungx:
Fix bug: when apply batch update or batch delete on a query with User Defined Table Type parameter, an exception will occur (System.ArgumentException : No mapping exists from object type System.Object[] to a known managed provider native type.).


borisdj:
Thx for contrib.

PR #456, opened 2021-01-29
Fix bulk delete with output table and add deletions to statistics

bennycoomans:
This PR fixes a bug where the following methods fail if an output table is used (e.g. `CalculateStats` is `true`):
- BulkDelete
- BulkDeleteAsync
- BulkInsertOrUpdateOrDelete
- BulkInsertOrUpdateOrDeleteAsync

The bug is caused by the fact that the `MERGE` statement does not output anything through the `INSERTED` collection for deleted records. Instead, it outputs those records in the `DELETED` collection. Therefore, when using `INSERTED`, an error about not being able to insert a `NULL` value is thrown.

I fixed the bug by using the `COALESCE` operator, which is (at least) available in MS Sql Server, Sqlite, PostgreSQL and MySql. When a value in the `INSERTED` collection is `NULL`, it falls back to the corresponding value in the `DELETED` collection.

Additionally, this PR fixes the `StatsInfo` so deletions are now counted separately in a `StatsNumberDeleted` property. Previously, deletions were counted under `StatsNumberInserted`.

Furthermore, this PR:
- Adds tests for InsertOrUpdateOrDelete(Async)
- Enables `CalculateStats` for all bulk operation tests that did not have it yet, so the fix is tested for those operations as well:
  - BulkInsert
  - BulkInsertAsync
  - BulkUpdate
  - BulkDelete
  - BulkDeleteAsync
- Validates the `StatsInfo` results in all relevant unit tests
- Adds a query and query component to `TestExample.sql` to make it incorporate the fix as well

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
@@ -168,7 +168,8 @@ namespace EFCore.BulkExtensions.Tests
                                     PreserveInsertOrder = true,
                                     SetOutputIdentity = true,
                                     BatchSize = 4000,
-                                    UseTempDB = true
+                                    UseTempDB = true,
+                                    CalculateStats = true
                                 },
                                 (a) => WriteProgress(a)
                             );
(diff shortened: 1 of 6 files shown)


borisdj:
Thx for contrib 👍

PR #457, opened 2021-02-01
Add an Activity to all operations.

mjeanrichard:
This PR instruments the Bulk Operations with Activties. This allows to collect telemetry about the bulk insert performance.


borisdj:
Thx for Contrib. 👍

PR #461, opened 2021-02-06
Fix line endings in EFCoreBatchTest.cs

bennycoomans:
The file `EFCoreBatchTest.cs` is the only file with CRLF, LF line terminators. This causes a major diff for minor changes. So in this PR, I only re-saved the file with consistent line endings, in order to be able to make future changes with a normal diff.

Before:

Screenshot 2021-02-06 at 08 06 31

After:

Screenshot 2021-02-06 at 08 07 12


borisdj:
Merged 👍

PR #462, opened 2021-02-07
Update TableInfo to not use GetMethod so it's compatible with EF Core 5

### LINKED ISSUE #443: EF Core 5 throws AmbiguousMatchException
I understand that EF Core 5 isn't supported yet - but I've tried using BulkExtensions with it and ran into a small issue where `DbContext` now has 2 overloads of `Set<T>`, which causes an `AmbiguousMatchException` in `TableInfo.GetQueryExpression`:
https://github.com/borisdj/EFCore.BulkExtensions/blob/ef1ce1d3416b197afd09d566b36a6b7621c35216/EFCore.BulkExtensions/TableInfo.cs#L723

My suggested fix is to use the overload of `Expression.Call` that takes a method name and generic types:
```cs
            var expression = Expression.Call(parameter, "Set", new Type[] { entityType });
```

The same change could be made for `FromSqlRaw` and `AsNoTracking`
https://github.com/borisdj/EFCore.BulkExtensions/blob/ef1ce1d3416b197afd09d566b36a6b7621c35216/EFCore.BulkExtensions/TableInfo.cs#L725
https://github.com/borisdj/EFCore.BulkExtensions/blob/ef1ce1d3416b197afd09d566b36a6b7621c35216/EFCore.BulkExtensions/TableInfo.cs#L732

I'm happy to submit a PR for this.

pmccloghrylaing:
- EF Core 5 has multiple DbContext.Set methods that cause GetMethod to throw an AmbiguousMatchException
- Use Expression.Call overload that takes instance, method name, generic types and arguments

Closes #443

### DIFF
--- a/EFCore.BulkExtensions/TableInfo.cs
+++ b/EFCore.BulkExtensions/TableInfo.cs
@@ -195,7 +195,7 @@ namespace EFCore.BulkExtensions
             {
                 if (property.PropertyInfo != null) // skip Shadow Property
                 {
-                    FastPropertyDict.Add(property.Name, new FastProperty(property.PropertyInfo));   
+                    FastPropertyDict.Add(property.Name, new FastProperty(property.PropertyInfo));
                 }
             }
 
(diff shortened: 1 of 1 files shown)


borisdj:
Thx for Contrib. 👍

PR #467, opened 2021-02-14
Add (normal) Shadow Property support when BulkConfig.EnableShadowProperties is true

### LINKED ISSUE #470: Getting exceptions related to (normal) shadow properties after EF Core 3.1 upgrade
Hi,

When doing a BulkInsertOrUpdateAsync or other bulk operations I'm getting an exception like that:
_**The given value 'Entity' of type String from the data source cannot be converted to type bit for Column 5 [IsDeleted] Row**_ 

I found a workaround for it to use a PropertiesToExclude in BulkConfig, however when I use it I'm getting a different exception:
_**Cannot insert the value NULL into column 'IsDeleted', table 'xxx'; column does not allow nulls. UPDATE fails.
The statement has been terminated.**_

I tried a suggested method to use Context.AddRange() and also tried to do something like that:
`Context.Entry(value).Property("IsDeleted").CurrentValue = false;`
However it still didn't work.

I wasn't getting those exceptions until I have upgraded EF Core to 3.1.x

Code:
```
                if (Context.ChangeTracker != null)
                {
                    Context.ChangeTracker.AutoDetectChangesEnabled = false;
                    var config = new BulkConfig()
                                 {
                                     PropertiesToExclude = new List<string>() { "IsDeleted" }
                                 };

                    await Context.BulkInsertOrUpdateAsync(values, config);
                }
```

maitlandmarshall:
Closes #470 

Add (normal) Shadow Property support when BulkConfig.EnableShadowProperties is true. You can now add Shadow Properties to your model and persist the values without an error.

```
using var db = new ExampleDbContext();
db.BulkInsertOrUpdate(this.GetTestData(db).ToList(), new BulkConfig
{
    EnableShadowProperties = true
});

private IEnumerable<SpModel> GetTestData(DbContext db)
{
    var one = new SpModel();
    db.Entry(one).Property(SpModel.SpLong).CurrentValue = (long)10;
    db.Entry(one).Property(SpModel.SpNullableLong).CurrentValue = null;
    db.Entry(one).Property(SpModel.SpDateTime).CurrentValue = new DateTime(2021, 02, 14);

    yield return one;
}
```

This disables the automatic discrimator feature, you will have to use manual model method instead.

Add testsettings.json with support for testsettings.local.json for those of us who don't have a local SQL server.

### DIFF
--- a/EFCore.BulkExtensions/EFCore.BulkExtensions.csproj
+++ b/EFCore.BulkExtensions/EFCore.BulkExtensions.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="3.1.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="3.1.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.1.8" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.12" />
     <PackageReference Include="System.Data.SqlClient" Version="4.6.0" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.1" />
   </ItemGroup>
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ obj
 .vs
 *.user
 .idea
+/EFCore.BulkExtensions.Tests/testsettings.local.json
--- a/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
+++ b/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
@@ -27,4 +27,13 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
 
+  <ItemGroup>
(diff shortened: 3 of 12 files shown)


jura1514:
Well done, I think that might be exactly what I need, recently opened an issue related to shadow properties
[https://github.com/borisdj/EFCore.BulkExtensions/issues/470](url)

borisdj:
Thx for contrib.

EFCore.BulkExtensions/EFCore.BulkExtensions.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="3.1.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="3.1.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.1.8" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.12" />
```
Numpsy, line 26: Is this needed in the main library if the ConfigurationBuilder stuff is only used in the test project?

borisdj: Moved to Test only

Issue #466, opened 2021-02-12
AmbiguousMatchException on update from 3.2.7 to 3.3.1 with EF core 5


jr01:
This line in [BatchUtil.cs](https://github.com/borisdj/EFCore.BulkExtensions/blob/master/EFCore.BulkExtensions/BatchUtil.cs#L413)
```
typeof(DbContext).GetMethod(nameof(DbContext.Set), BindingFlags.Public | BindingFlags.Instance);
```                

throws `AmbiguousMatchException`.

In EF core 5 there are 2 set methods:
```
public virtual DbSet<TEntity> Set<TEntity>() where TEntity : class
public virtual DbSet<TEntity> Set<TEntity>(string name) where TEntity : class
```
See https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.set?view=efcore-5.0#Microsoft_EntityFrameworkCore_DbContext_Set__1_System_String_

I think a `typeof(DbContext).GetMethod(nameof(DbContext.Set), new Type[] {});` should fix it.

JavidShoaei:
Hello @jr01 Did you maybe fix it yourself? If so, would you please open a merge request? I'm in desperate need.

jr01:
@JavidShoaei - If I had a big need I would put my time into a PR ;) The fix for this is probably easy, but testing...

ryanlangton:
https://github.com/borisdj/EFCore.BulkExtensions/pull/475

borisdj:
Fix published 3.3.2

PR #477, opened 2021-03-04
#466 Fix AmbiguousMatchException on update from 3.2.7 to 3.3.1 with EF core 5

JavidShoaei:
Fixes the issue with EF Core 5 AmbiguousMatchException.
#466


JavidShoaei:
Fixes the issue with EF Core 5 `AmbiguousMatchException`.
#466

borisdj:
There was a prior commit https://github.com/borisdj/EFCore.BulkExtensions/pull/475 for this fix:
`.GetMethod(nameof(DbContext.Set), new Type[] {});`

JavidShoaei:
I think using `Array.Empty` instead of `new Type[]` is slightly better though.

borisdj:
Changed.

PR #478, opened 2021-03-06
Add option to exclude properties from the insert and/or update portions in BulkInsertOrUpdate method.

codecooper:
Add option to exclude properties from the insert and/or update portions in BulkInsertOrUpdate method.

Let's say we have CreationDate/CreationBy properties - it should be inserted but never updated. We also have UpdateDate/UpdateBy properties - It should update when the row is updated but not because those values were updated. The four new properties **PropertiesToIncludeOnCompare**, **PropertiesToExcludeOnCompare**, **PropertiesToIncludeOnUpdate**, and **PropertiesToExcludeOnUpdate** allow for those columns to be specified. An added benefit of these new features is that even if they are not explicitly implimented by the user, *MERGE* performance is greatly improved. One of the slowest operations on a database is writing. The *EXISTS / EXCEPT* clause on the *MERGE* command will now only update rows that have changes instead of all records that have a match in the temp table.

### DIFF
--- a/EFCore.BulkExtensions.Tests/SqlQueryBuilderTests.cs
+++ b/EFCore.BulkExtensions.Tests/SqlQueryBuilderTests.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 
 namespace EFCore.BulkExtensions.Tests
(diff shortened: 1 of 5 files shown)


borisdj:
Thx for contrib.

jr01:
Nice except this is wrong:
```
        bool AreSpecifiedPropertiesToIncludeOnUpdate = BulkConfig.PropertiesToExcludeOnUpdate?.Count() > 0;
            bool AreSpecifiedPropertiesToExcludeOnUpdate = BulkConfig.PropertiesToIncludeOnUpdate?.Count() > 0;
```

should be
```
```
        bool AreSpecifiedPropertiesToIncludeOnUpdate = BulkConfig.PropertiesToIncludeOnUpdate?.Count() > 0;
            bool AreSpecifiedPropertiesToExcludeOnUpdate = BulkConfig.PropertiesToExcludeOnUpdate?.Count() > 0;
```

borisdj:
@jr01  Fixed 👍

borisdj:
This config is not default, it is opt-in, so no need to opt-out, just don't set it.

macwier:
@borisdj 
Nope, it is enabled by default. Quoting the OP:
*An added benefit of these new features is that **even if they are not explicitly implimented by the user**, MERGE performance is greatly improved. One of the slowest operations on a database is writing. The EXISTS / EXCEPT clause on the MERGE command will now only update rows that have changes instead of all records that have a match in the temp table.*

Looking at the code, the only way to disable this is to have a spatial column in the table: 
https://github.com/borisdj/EFCore.BulkExtensions/blob/f69b6d034d89df5fb5fb7fdb229ab9e10ac6671c/EFCore.BulkExtensions/SqlQueryBuilder.cs#L191-L196

borisdj:
I thought you only meant about new configs Properties-ToInclude/ToExclude-OnCompare/OnUpdate which are empty by default, but are referring to sql `EXCEPT` code. I'll add config for disabling this segment.

borisdj:
Added with v3.4.5

borisdj:
_SkipClauseExistsExcept_ is renamed to _**OmitClauseExistsExcept**_ with v3.5.6

Brondahl:
I note that this only supports SqlServer?
Is it feasible to extend this to other Sql flavours?

If not, is there scope for validating the config and rejecting attempts to configure these options in a non-SqlServer context?

Issue #399, opened 2020-10-01
BatchDelete fails on queries that use EFCore Include


metoule:
Let's say I have the following entities:

````C#
class RockBand {
    [Key]
    public int RockBandID { get; set; }
    public List<Member> Members { get; set; }
}

class Member {
    [ForeignKey(nameof(RockBand ))]
    public int RockBandID { get; set; }
    public RockBand RockBand { get; set; }
}
````

I want to delete all `RockBand` entities with no `Member`, so I wrote the following statement:

````
_context.RockBands.Include(_ => _.Members).Where(_ => !_.Members.Any()).BatchDelete();
````

This call fails with the exception:

> Incorrect syntax near the keyword 'ORDER'.

The generated SQL statement is:

````
DELETE [r]
FROM [RockBand] AS [r]
LEFT JOIN [Member] AS [r0] ON [r].[RockBandID] = [r0].[RockBandID]
WHERE NOT (EXISTS (
    SELECT 1
    FROM [Member] AS [r1]
    WHERE [r].[RockBandID] = [r1].[RockBandID]))
ORDER BY [r].[RockBandID], [r0].[MemberID]
````

The issue comes from the fact that EF Core always appends an `ORDER BY` clause when joining multiple tables.
Is there a way for BulkExtensions to strip this `ORDER BY` clause that causes the DELETE to fail?

Issue #301, opened 2020-02-28
Incorrect syntax near the keyword 'ORDER'.


tomasr78:
The query below produce SQL error

```
Microsoft.Data.SqlClient.SqlException : Incorrect syntax near the keyword 'ORDER'.
Data:
  HelpLink.ProdName: Microsoft SQL Server
  HelpLink.ProdVer: 11.00.7001
  HelpLink.EvtSrc: MSSQLServer
  HelpLink.EvtID: 156
  HelpLink.BaseHelpUrl: http://go.microsoft.com/fwlink
  HelpLink.LinkId: 20476
```

`_db.Logs.Where(w => w.DateStamp < DateTime.Now.AddYears(-1)).OrderBy(o => o.DateStamp).Take(take).BatchDeleteAsync(cancellationToken)`

borisdj:
DELETE with TOP and ORDER BY requries [**Subselect** statement](https://sqlstudies.com/2018/09/19/you-cant-delete-top-x-with-an-order-by/), and Batch uses simple query conversion which has no subselect.

jr01:
I use the following workaround for this issue.

```c#
    public static class MyQueryableBatchExtensions
    {
        public static async Task<int> BatchDeleteAsync_cte(this IQueryable query, CancellationToken cancellationToken = default)
        {
            DbContext context = BatchUtil.GetDbContext(query);
            (string sql, var sqlParameters) = GetSqlDelete_cte(query, context);
            return await context.Database.ExecuteSqlRawAsync(sql, sqlParameters, cancellationToken).ConfigureAwait(false);
        }

        private static (string, List<object>) GetSqlDelete_cte(IQueryable query, DbContext context)
        {
            (string sql, string _, string _, string topStatement, string _, IEnumerable<object> innerParameters) = BatchUtil.GetBatchSql(query, context, isUpdate: false);

            innerParameters = BatchUtil.ReloadSqlParameters(context, innerParameters.ToList()); // Sqlite requires SqliteParameters
            var resultQuery = $"WITH MyCTE AS (SELECT {topStatement + "*" + sql}) DELETE FROM MyCTE";
            return (resultQuery, new List<object>(innerParameters));
        }
    }
```

I guess it also works for https://github.com/borisdj/EFCore.BulkExtensions/issues/399

@borisdj - if you want I can make a PR.

PR #483, opened 2021-03-13
MSSQL batch delete with order by queries

jr01:
1. Use an updateable CTE in DELETE when the query uses TOP/OFFSET with ORDER BY
2. Strip ORDER BY in DELETE when the query does not use TOP/OFFSET

Fixes https://github.com/borisdj/EFCore.BulkExtensions/issues/301
Fixes https://github.com/borisdj/EFCore.BulkExtensions/issues/399

Notes
* Why not use an updateable CTE for all queries? 
   a. Updateable CTE has limitations like only a single main table (no JOINs).
   b. Updateable CTE still doesn't allow ORDER without TOP/OFFSET, so stripping ORDER BY (#2) is needed anyway.
   
* I didn't add support for SQLite
  SQLite does not support updateable CTE, it does support a normal CTE and I got something like this working:
  ```
  var entityType = context.Model.FindEntityType(query.ElementType);
  var pkeyExpression = entityType.FindPrimaryKey();	
  var pkey = this.Escape(pkeyExpression.Properties[0].GetColumnName());
	
  $"WITH cte AS (SELECT {tbl}.{pkey} {sql} DELETE FROM {tbl} AS cte_del WHERE dte_del.{pkey} IN cte"
  ```
  
  But this approach has some limitations:
  - EF supports entities without primary keys...but the subquery needs a unique column...
  - EF supports entities with compound primary keys and WHERE IN only supports a single column...
  
  In the end I decided getting this working in SQLite wasn't worth my time.

* I did my changes in `BatchUtil.cs` and not in the (new) `SqlServerDialect.cs`
  I think the SqlServerDialect should have a method like: `GetSqlDelete(IQueryable query, DbContext context)` and all the complexity is taken out of BatchUtil, but that requires a lot of refactoring and is an exercise for perhaps another time.
  			
* I didn't build support for these kinds of queries in batch UPDATE

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBatchTest.cs
@@ -37,7 +37,7 @@ namespace EFCore.BulkExtensions.Tests
             RunContainsBatchDelete2();
             RunContainsBatchDelete3();
             RunAnyBatchDelete();
-
+            
             using (var context = new TestContext(ContextUtil.GetOptions()))
             {
                 var firstItem = context.Items.ToList().FirstOrDefault();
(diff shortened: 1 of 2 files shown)


borisdj:
Thx for contrib.

PR #487, opened 2021-03-15
Add ability to insert, update or merge the entity object graph using BulkConfig.IncludeGraph = true

maitlandmarshall:
Insert, update or merge the entire entity object graph. Does not support Sqlite due to how its set output identity = true implementation works.

```csharp
private static WorkOrder WorkOrder1 = new WorkOrder
{
    Description = "Fix belt",
    Asset = new Asset
    {
        Description = "MANU-1",
        Location = "WAREHOUSE-1"
    },
    WorkOrderSpares =
        {
            new WorkOrderSpare
            {
                Description = "Bolt 5mm x5",
                Quantity = 5,
                Spare = new Spare
                {
                    PartNumber = "MZD 5mm",
                    Barcode = "12345"
                }
            },
            new WorkOrderSpare
            {
                Description = "Bolt 10mm x5",
                Quantity = 5,
                Spare = new Spare
                {
                    PartNumber = "MZD 10mm",
                    Barcode = "222655"
                }
            }
        }
};

private IEnumerable<WorkOrder> GetTestData(DbContext db)
{
    yield return WorkOrder1;
}

var testData = this.GetTestData(db).ToList();
await db.BulkInsertOrUpdateAsync(testData, new BulkConfig
{
    IncludeGraph = true
});
```

### DIFF
--- a/EFCore.BulkExtensions.Tests/IncludeGraph/GraphDbContext.cs
+++ b/EFCore.BulkExtensions.Tests/IncludeGraph/GraphDbContext.cs
@@ -31,15 +31,15 @@ namespace EFCore.BulkExtensions.Tests.ShadowProperties
                 cfg.HasKey(y => y.Id);
 
                 cfg.HasMany(y => y.WorkOrderSpares).WithOne(y => y.WorkOrder);
-                cfg.HasOne(y => y.Asset).WithMany();
+                cfg.HasOne(y => y.Asset).WithMany().IsRequired();
             });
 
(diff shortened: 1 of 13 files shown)


borisdj:
Thx for contrib.
Btw I am currently working on BulkSaveChanges ops that will have similar capability.
Automatically Insert/Update-ing to multiple tables.
Difference being that for BulkSaveChanges will work on all chaned entites(the don't have to connected) but they will have to be added to DbSet where they are tracked.
With Graph option that is here implemented there is no need for tracking them.

PR #494, opened 2021-03-19
Adds support for ValueConverter in BatchUpdate Expression

jr01:
This fixes an issue where `BatchUpdate` with an Expression would write the unconverted value to the DB, instead of using the configured ValueConverter.

For example with an Enum property that has a string conversion this would write an integer to the DB instead of a string.

```c#
public enum VcEnum
{
    Why,
    ...
}

cfg.Property(y => y.Enum).HasColumnType("nvarchar(4000)").HasConversion<string>();

...
db.VcModels.BatchUpdate(x => new VcModel
{
    Enum = VcEnum.Why
});
```          

wrote `0` to the DB instead of `"Why"`


borisdj:
Thx for contrib.

PR #495, opened 2021-03-19
BulkInsert problem when primary key has be defined ColumnName and SetOutputIdentity is true.

hhendrikk:
BulkInsert problem when primary key has be defined ColumnName and SetOutputIdentity is true.

Problem:

Mapping sample:

        builderModel.ToTable("TableSample", "dbo");

        builderModel.Property(x => x.Id)
            .HasColumnName("ID_TS")
            .IsRequired();

        builderModel.HasKey(x => x.Id);
        .
        .
        . 
Select generated on TempTable:

SELECT [r].[ID_TS], ...
FROM (
SELECT [ID_TS], ... FROM [dbo].[TableSample431421f7Output] WHERE **_[Id]_** IS NOT NULL
) AS [r]
ORDER BY [r].[ID_TS]

Generated "where" with property name than [ID_TS] (property column name).


borisdj:
Thx for contrib.

Issue #397, opened 2020-09-28
EnumToString not working in BatchUpdateAsync


IeuanWalker:
Hi,

I have this table - 
```csharp
[Table("Settings")]
public class SettingsTbl
{
    [Key]
    public SettingsEnum Setting { get; set; }
    [Required]
    [MaxLength(20)]
    public string Value { get; set; }
}
```
and this in the database context - 
```csharp
modelBuilder
    .Entity<SettingsTbl>()
    .Property(e => e.Setting)
    .HasConversion<string>();
```

Calling this method -
```csharp
public async Task UpdateSetting(SettingsEnum setting, object value)
{
    await dbSet
        .Where(x => x.Setting == setting)
        .BatchUpdateAsync(x => new SettingsTbl
        {
            Value = value.ToString()
        })
        .ConfigureAwait(false);
}
```

Throws the following error - 
![image](https://user-images.githubusercontent.com/6544051/94444888-b64ea500-019e-11eb-9582-6fd7c13aea50.png)

Using NuGet version - 3.2.3

borisdj:
Conversion is not currently supported for Batch.

IeuanWalker:
@borisdj thanks, is this being planned?

Not too familiar how this package works, but if I figure out how to add this feature ill create a PR for it.

borisdj:
If you manage to do it, you can send PR.

borisdj:
Solved with latest version.
However for it to work name of parameter has to same as Property name (Case-insensitive), example:
_context.Settings.Where(x => x.**Settings** == **settings**).BatchUpdate(x => new Setting { Value = value.ToString() });_

borisdj:
As of Now it works even with different named params.

PR #270, opened 2019-12-11
BatchDelete works with EFCore 3.1 for SQL Server

halllo:
The reflection in IQueryableExtensions didn't work for EFCore 3.1. What do you think?

PR #497, opened 2021-03-22
Fix BatchDelete and BatchUpdate with where expression and ValueConverter

jr01:
Hello again :)

When using something like:
```c#
var date = new LocalDate(2020, 3, 21);
dbContext.Model.Where(x => x.LocalDate > date).BatchDelete()
```

I got this exception:
```c#
System.ArgumentException: No mapping exists from object type EFCore.BulkExtensions.Tests.ValueConverters.LocalDate to a known managed provider native type.
	at Microsoft.Data.SqlClient.MetaType.GetMetaTypeFromValue(Type dataType, Object value, Boolean inferLen, Boolean streamAllowed)
	at Microsoft.Data.SqlClient.MetaType.GetMetaTypeFromType(Type dataType)
	at Microsoft.Data.SqlClient.SqlParameter.GetMetaTypeOnly()
	at Microsoft.Data.SqlClient.SqlParameter.get_DbType()
	at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlServerDialect.ReloadSqlParameters(DbContext context, List`1 sqlParameters)
	at EFCore.BulkExtensions.BatchUtil.ReloadSqlParameters(DbContext context, List`1 sqlParameters)
	at EFCore.BulkExtensions.BatchUtil.GetSqlDelete(IQueryable query, DbContext context)
	...
```

The problem was that `CheckAndSetParametersForConvertibles` only applied the converter when `if (convertibleProperty.Key.ToLower() == parameterColumnName)` and that isn't the case here: `convertibleProperty.Key = 'LocalDate'` and `parameterColumnName = 'date'`.  And also the `CheckAndSetParametersForConvertibles` only worked when dealing with a string database type.

The solution is to let EF Core convert the parameter itself using `IRelationalParameter.AddDbParameter(dbCommand, values);` in `ToParametrizedSql()`.
	
I'm not sure if this fix also needs to be in the else part of `ToParametrizedSql()` when `relationalCommandCache == null`, because I have no clue how to trigger that branch with EF Core 3.1 ...

### DIFF
--- a/EFCore.BulkExtensions.Tests/ValueConverters/ValueConverterTests.cs
+++ b/EFCore.BulkExtensions.Tests/ValueConverters/ValueConverterTests.cs
@@ -20,7 +20,7 @@ namespace EFCore.BulkExtensions.Tests.ValueConverters
 
             using var db = new VcDbContext(ContextUtil.GetOptions<VcDbContext>(databaseName: $"{nameof(EFCoreBulkTest)}_ValueConverters"));
 
-            db.BulkInsertOrUpdate(this.GetTestData(db).ToList());
+            db.BulkInsertOrUpdate(this.GetTestData().ToList());
 
             var connection = db.Database.GetDbConnection();
             connection.Open();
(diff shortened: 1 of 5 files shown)


borisdj:
That was how it work, param name had to be same as of property: https://github.com/borisdj/EFCore.BulkExtensions/issues/397#issuecomment-798791339
But it was not best solution, more as a temporary.
I've moved setting params below _if-else_ segment so that it works for both.
Thx for the fix, v3.4.8 published.

Also I am not sure how to test `relationalCommandCache` being null.
It was a PR https://github.com/borisdj/EFCore.BulkExtensions/pull/270 from @halllo Do you know when will [relationalCommandCache = null](https://github.com/borisdj/EFCore.BulkExtensions/blob/3c40614b1eaf4868ce29783dbceff5b265d85e38/EFCore.BulkExtensions/IQueryableExtensions.cs#L33-L44) to make a test it?

Issue #321, opened 2020-04-16
Insert only if the entity does not already exists


API-kernel:
I want to insert entities but only those that don't already exist in the database.
I expect property in BulkConfig 
InsertOnlyIfNotExists - let you insert only entities that doesn't already exists.

in addition, I want the method to return the number of affected rows, it will be useful

ugumba:
Same here!  Something to the effect of sqlite's INSERT ON CONFLICT IGNORE would be great. Any solution I've seen without custom SQL suffers from race conditions and performance.problems.

borisdj:
At the moment you can use BulkInsertOrUpdate if they existing ones are the same, or use BulkRead and then remove those that exist from List before calling BulkInsert.

Masus:
Just to add to what @borisdj mentioned. You can get  a InsertIfNotExist behaviour by using the PropertiesToInclude on BulkConfig. You have to add something to it (empty list seems to be ignored) but just add any of the PK (which always will be the same).

This may affect timestamp columns or triggers but will get the IfNotExist behaviour without having to do a BatchRead and calculate it manually.

timothydowd:
Any chance you can elaborate on this? - i thought PropertiesToInclude just meant that only certain columns get added to a record, leaving the rest null - I don't understand how this can be used to ensure only records that are unique are added to the db...

borisdj:
Now it can be configured with separate PropertiesTo-Include/Exclude-OnUpdate.
More info in the README:

> If we want only to Insert new ones and to skip those that exist in Db then we can use BulkInsertOrUpdate with config `PropertiesToIncludeOnUpdate = new List<string> { "" }`

This was Dupe of: https://github.com/borisdj/EFCore.BulkExtensions/issues/189

Masus:
@timothydowd : Sorry for the very late reply. You are correct. I could use it as a InsertIfNotExist only because all the default values was actually the ones I wanted. Sorry for the confusion.

raysefo:
Any code sample can you share for inserting only if records do not already exist in the database?

StanlyLife:
For me `PropertiesToIncludeOnUpdate` did not work: 

```cs
        public async void AddRange(List<X> aci)
        {
            var bulkConfig = new BulkConfig() { PropertiesToIncludeOnUpdate = new List<string> { "" }, PropertiesToIncludeOnCompare = new List<string> { "Unixtime", "Time", "Activity" } };
[...]

Issue #189, opened 2019-05-20
RFI: Insert if missing, otherwise do nothing.


kellcomnet2:
Is there a way to only insert new rows based on the PK otherwise do nothing?

borisdj:
You can use BulkRead to get those that already exists in Db and exclude them from Insert.

pjh1974:
@borisdj but using 2 round-trips would not guard against the data changing between the BulkRead and the BulkInsert.  It would be helpful to have a BulkInsertNotExists or similar method that used a MERGE statement to INSERT WHEN MATCHED and do nothing WHEN NOT MATCHED.

borisdj:
What is your UseCase scenario for this?
Usually InsertOrUpdate is enough for most cases, and even if doing 2 queries to DB as mentioned it is still done relatively fast so small chance for conflict. Since you do not want to update existing only issue could be if some row would be deleted in between queries and then it not being inserted again.
Is this case likely to happen?

kellcomnet2:
In my case I insert the record with a Created_DT column with the update it overwrites the date each time.

borisdj:
Even then BulkRead can be used to set Created_DT column to value from DB so that it remains the same.

pjh1974:
Take the scenario where a process outputs a number of items which it persists to the database, these items can then be modified by the user.  Then, when the process is re-run it generates some of the old items again and also generates new items.  In order to create the new items and not overwrite/update the existing items (which may have user edits), I need to persist the new set, ignoring any rows that already exist.  

As you say this could be achieved by doing a BulkRead followed by a BulkInsert of only the items that were not returned by the BulkRead.  However, any new items added by another process between the calls to BulkRead and BulkInsert would cause the BulkInsert to fail due to a duplicate key violation, if the item happened to be a duplicate of one in the set to update.  The only solution would be to re-run the whole BulkRead, BulkInsert process again; which with large data sets would be costly.  

If I were solving this problem directly in SQL, I would implement it using a MERGE statement or using several INSERT statements, each with a where not exists check.

ralefire:
I would generalize the issue to include control over the merge operation "WHEN NOT MATCHED BY TARGET," "WHEN MATCHED," and "WHEN NOT MATCHED BY SOURCE." In my case I want to soft delete rows by updating a database column instead of actually deleting on the merge command "WHEN NOT MATCHED BY SOURCE." A fluent API for strongly typed objects that controls the merge operation would be excellent.

[...]

PR #501, opened 2021-03-23
InsertIfNotExists methods.

Yustos:
Great library.
I'm trying to add InsertIfNotExists(+async) method. Because trick with:
`PropertiesToIncludeOnUpdate = new List<string> { "" }`
is not working now (exception about unknown property is thrown and empty property list works incorrectly).
This helper can help for issues likes those:
#321 
#189 

I have no idea, how to write tests for this.
So i check this patch only on SQL server in my work projects.
SQLLite i didn't touch.

Thank you for this library!

### DIFF
--- a/EFCore.BulkExtensions/DbContextBulkExtensions.cs
+++ b/EFCore.BulkExtensions/DbContextBulkExtensions.cs
@@ -57,6 +57,30 @@ namespace EFCore.BulkExtensions
             DbContextBulkTransaction.Execute(context, entityType, entities, OperationType.InsertOrUpdate, bulkConfig, progress);
         }
 
+        public static void BulkInsertIfNotExists<T>(this DbContext context, IList<T> entities, BulkConfig bulkConfig = null, Action<decimal> progress = null) where T : class
+        {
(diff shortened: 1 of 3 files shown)


borisdj:
No problem, glad it's useful.
Have you tried it with latest version, empty list had bug that was fixed in 4.3.9
If it still doesn't work, write here code snipped with example where the bug can be reproduced.
I wouldn't like to make duplicate methods, since this feature can be configured.

Yustos:
Acknowledge. Now trick with empty property name:
```
await dbContext.BulkInsertOrUpdateAsync(entities.ToList(), new BulkConfig
{
	PropertiesToIncludeOnUpdate = new List<string> { "" },
```
works correct. And produce next statement:
```
MERGE [dbo].[MyTable] AS T
USING (SELECT TOP 43 * FROM [dbo].[MyTableTemp8a79dc95] ORDER BY [Id]) AS S
	ON T.[Id] = S.[Id]
WHEN NOT MATCHED BY TARGET THEN
	INSERT ([Field1], [Field2])
	VALUES (S.[Field1], S.[Field2]);
```

Thank you!

borisdj:
Great.

Issue #534, opened 2021-04-22
SetOutputIdentity = true does not always work in BulkUpdateOrInsert


huer12:
If the entity list has entities to update and entities to insert then the following code in the method TableInfo.UpdateEntitiesIdentity() prevents that the Id is updated:
`if (countDiff > 0) // When some ommited from Merge because of TimeStamp conflict then changes are not loaded but output is set in TimeStampInfo`

If the row to update has not changed, then the SQL merge will not update this row and the count will be different. Because of that countDiff is bigger than 0 and as a result the ids of the inserted rows are not updated.

You could argue, that you shouldn't try to update rows which haven't changed. Well that's true, but sometimes it is just easier to update these rows anyway. 
Furthermore it is a really nasty bug, which occurred in production, not in the test/dev environment. And the error message a FK exception in the dependent class which we tried to update. 

I tried the workaround with PreserveInsertOrder = false, as I have seen, that the list gets cleared and added with the entities from the temp table. Also stated in the documentation:
_When used Id-s will be updated in entitiesList, and if PreserveInsertOrder is set to false then entitiesList will be cleared and reloaded._

But the reloaded list is never returned. (Reason is Line 782 TableInfo.LoadOutputDataAsync) is that by purpose or a bug?

huer12:
The simplest workaround for us would be to perform the update and insert in to steps. But this doesn't work with UniqueTableNameTempDb = false within the same transaction. Would it be an option for you to extend the BulkConfig with the Porperty "TempTableName" (or a prefix or suffix)? Then the BulkUpdate and BulkInsert could be provided with different names.

(UniqueTableNameTempDb = false is a requirement from our DB engineers so the SQL server does not always re-compile)

borisdj:
You can make _countDiff_ to be zero, with BulkConfig setting `OmitClauseExistsExcept` to **_True_**, so that even same data is updated.

As for the loaded list, with Omit you can get regular reload of properties.
But even when `PreserveInsertOrder` is _false_ loaded list is not explicitly returned from method but it is still accessible via reference of entities List that was sent as argument in Bulk operation since that list is cleared and filled with loaded entities.

Also when _countDiff > 0_ EntitiesOutput is returned in `BulkConfig.TimeStampInfo.EntitiesOutput.` 

So in conclusion there is no need for extra config.

huer12:
Hi
Thanks for the explanation!
Regarding this point:
[...]

PR #536, opened 2021-04-23
Add possibility to configure the suffix of the temp table name

huer12:
Add possibility to configure the suffix of the temp table name via BulkConfig


borisdj:
Extra config not required: https://github.com/borisdj/EFCore.BulkExtensions/issues/534

PR #545, opened 2021-05-01
Adds optional delete filter to BulkInsertOrUpdateOrDelete

jr01:
Hi,

I have a muli-tenant system where tenants share the same table using a TenantId column. And I would like to be able to do a bulk merge with data for 1 specific tenant using `BulkInsertOrUpdateOrDelete`. The entities to insert and update have a TenantId property set, so that's ok, but obviously that can't be done for delete and currently `BulkInsertOrUpdateOrDelete` removes data from all tenants.

In SQL it's possible to do:`MERGE .... WHEN NOT MATCHED BY SOURCE AND [TenantId] = @currentTenantId THEN DELETE`
(https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15).

This can now be done through an optional parameter that takes an expression:
```c#
context.BulkInsertOrUpdateOrDelete(entities, bulkConfig, deleteFilter: e => e.TenantId == currentTenantId);
```

Let me know what you think! And I'm happy to change things of course.

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
@@ -269,13 +269,18 @@ namespace EFCore.BulkExtensions.Tests
                     TimeUpdated = dateTimeNow
                 });
             }
+
+            int? keepEntityItemId = null;
             if (isBulk)
             {
                 var bulkConfig = new BulkConfig() { SetOutputIdentity = true, CalculateStats = true };
-                context.BulkInsertOrUpdateOrDelete(entities, bulkConfig, (a) => WriteProgress(a));
+
+                keepEntityItemId = 3;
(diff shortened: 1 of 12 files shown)


borisdj:
Nice work, I've scanned the PR and all seems good.
Just one thing, maybe _deleteFilter_ instead of extra argument in method, to be in BulkConfig as Property with name **SynchronizeFilter** since _Synchronize_ is alias for _BulkInsertOrUpdateOrDelete_.
That way no need to change method signatures and they all remain with same args.
What do you think?

jr01:
Good idea @borisdj ! I just applied your suggestion. 
One small note: I didn't want to change `BulkConfig` to `BulkConfig<T> where T: class`, so that's why it's a `SetSynchronizeFilter<T>` method.

borisdj:
Published with 5.1.4
Thx for contrib.

PR #575, opened 2021-06-18
Use SQLitePCLRaw.bundle_e_sqlite3 instead of SQLitePCLRaw.bundle_winsqlite3

0xced:
So that the SQLite tests pass also on Linux and macOS

Otherwise it fails with a System.DllNotFoundException:
> Unable to load shared library 'winsqlite3' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libwinsqlite3, 1): image not found

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
+++ b/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
@@ -8,7 +8,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.*" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer.NetTopologySuite" Version="5.0.*" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="5.0.0" />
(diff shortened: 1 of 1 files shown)


palhal:
Using _bundle_winsqlite3_ causes problems on Windows Server 2019 also. I just spent a couple of hours debugging the following exception in my production environment:

> Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'near "ON": syntax error'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.PrepareAndEnumerateStatements(Stopwatch timer)+MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.Prepare()
   at EFCore.BulkExtensions.SQLAdapters.SQLite.SqLiteOperationsAdapter.GetSqliteCommand[T](DbContext context, Type type, IList`1 entities, TableInfo tableInfo, SqliteConnection connection, SqliteTransaction transaction)

It turns out Windows Server 2019 has an older version (3.23) of _winsqlite3.dll_ than Windows 10 (3.29). Not sure if that is the only reason though. Using _bundle_e_sqlite3_ instead fixed the problem.

PR #587, opened 2021-07-08
Include XML documentation in the NuGet package.

DarkDaskin:
Some members have XML documentation in code, but it is not included into produced NuGet packages. This PR fixes it.


borisdj:
Thx for contrib.
v5.2.7 published.

PR #588, opened 2021-07-08
If the input have a value for property that have HasDefaultValueSql, don't skip this property in insert state

ferndem:
If the input have a value for property that have HasDefaultValueSql, don't skip this property in insert state.


borisdj:
v5.2.8 published
Thx for contrib.

PR #659, opened 2021-11-11
Use column conversions for default values when generating sql SET segment

### LINKED ISSUE #658: BatchUpdate does not consider conversions for default values
How to reproduce:
* Configure a DateTimeOffset property called X on your entity to use a Conversion (e.g. DateTimeOffsetToBinaryConverter)
* Add an instance of your entity with a value '2021-11-11' for X, save the db context
* Prepare another instance of your entity, don't set X explicitly (will remain at DateTimeOffset.MinValue)
* Call BatchUpdateAsync, the overload with argument "object updateValues", and pass in your instance without any values set

Current behavior:
* The existing entity in the db context will get its property X overwritten to '0001-01-01'.

Expected behavior:
* The existing entity in the db context will keep its previous X value of  '2021-11-11'.

fdahlen:
Fixes #658

### DIFF
--- a/EFCore.BulkExtensions.Tests/BatchUtil/BatchUtilTests.cs
+++ b/EFCore.BulkExtensions.Tests/BatchUtil/BatchUtilTests.cs
@@ -1,7 +1,7 @@
 ﻿using EFCore.BulkExtensions.SqlAdapters;
 using Xunit;
 
-namespace EFCore.BulkExtensions.Tests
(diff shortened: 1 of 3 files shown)


fdahlen:
Includes unit test

fdahlen:
@borisdj Thanks for accepting. Would it be possible to get this fix also for EFCore5.0? Will take some time before we (and I suspect everyone else) migrate to 6.0.

Should I create another PR for that?

PR #668, opened 2021-11-29
DefiningNavigationName and DefiningEntityType are not supported anymore 

davidstjacques:
They always return null
Same fix as in TableInfo.cs


borisdj:
👍

Issue #67, opened 2018-08-12
Postgresql Integrations


nsunmaz:
I use "**Npgsql.EntityFrameworkCore.PostgreSQL**" with EF Core in the project. Does your library support **PostgreSQL**? If not, do you have a plan to do this later?

neco122:
Thanks for information.

collinbarrett:
Looks like a great library! I am also not using SQL Server, though. I'd recommend making the db provider support more prominent on the README, as it is not immediately obvious to potential users that it only supports SQL Server.

borisdj:
It is stated at the end of first paragraph:

> Under the hood uses SqlBulkCopy for Insert, for Update/Delete combines BulkInsert with raw Sql MERGE (MsSQL 2008+).

richard-s-popov:
Still no news about support PostgreSQL?

borisdj:
Not yet.
Have managed to extend it for SQLite recently.
PostgreSql maybe sometimes during next year.

MaceWindu:
As workaround you can use linq2db [BulkCopy](https://github.com/linq2db/linq2db#bulk-copy) functionality through [EF.Core adapter](https://github.com/linq2db/linq2db.EntityFrameworkCore)

https://github.com/linq2db/linq2db/wiki/Bulk-Copy#postgresql-provider-specific-bulk-copy

IvanAntipov:
EFCore.BuildExtension is a great library, unfortunately it doesn't support Postgres. You can use 'FlexLabs.Upsert' nuget as alternative for Postgres

etiennecl:
Is this still in the roadmap? Anything I could do to help? I'm not an expert but I sure could give you a hand.

borisdj:
Yes but still no timeframe.
There is one PR a few days ago in this direction: https://github.com/borisdj/EFCore.BulkExtensions/pull/434

Here are some links (Docs) that I've found when did the research:

https://www.npgsql.org/doc/copy.html
https://www.postgresql.org/docs/9.2/static/sql-copy.html
https://github.com/npgsql/Npgsql.EntityFrameworkCore.PostgreSQL/issues/113
__
https://www.bytefish.de/blog/postgresql_bulk_insert/
https://dba.stackexchange.com/questions/41059/optimizing-bulk-update-performance-in-postgresql
http://thebuild.com/blog/2016/01/14/doing-a-bulk-merge-in-postgresql-9-5/
[...]

PR #677, opened 2021-12-04
Fix Postgres BulkInsert CLR type matching

Hoffs:
Instead of `Write`'ing just value also provide column type name when
available. In this case correct column type will be used for CLR
types that are not natively supported or mapped. While this pretty
much fixes the issue #67 (specifically https://github.com/borisdj/EFCore.BulkExtensions/issues/67#issuecomment-982876139 ) , relying on final column
type can lead to issues when assumed type handler doesn't match the actual type.
For example `uint32` has pg type `bigint`, but default handler for this
type is `Int64Handler` which doesn't handle `uint32`.

But regardless of that, this should fix majority of issues with CLR type mapping.

Also I am not entirely sure whats the pattern with using non-async variants inside
async functions. Especially when nested non-async variants call async variants
with `GetAwaiter().GetResult()`. Seems pretty wasteful and not very performant.
I noticed somewhat similar pattern in Npgsql code, so maybe it makes some sense.
But regardless of that, I found that when using non-async variants in `BulkInsert`,
running 4 tasks inserting 1 million entities each (50k batches), frequently only single
task is started at a time. Changing to async in b36d013dee8d0a9fe50a696f6fd477fdd77a8759 alleviates this and
shows each task starting around same time.

### DIFF
--- a/EFCore.BulkExtensions/SQLAdapters/PostgreSql/PostgreSqlAdapter.cs
+++ b/EFCore.BulkExtensions/SQLAdapters/PostgreSql/PostgreSqlAdapter.cs
@@ -41,7 +41,7 @@ namespace EFCore.BulkExtensions.SQLAdapters.PostgreSql
             {
                 string sqlCopy = SqlQueryBuilderPostgreSql.InsertIntoTable(tableInfo, tableInfo.InsertToTempTable ? OperationType.InsertOrUpdate : OperationType.Insert);
 
-                using var writer = connection.BeginBinaryImport(sqlCopy);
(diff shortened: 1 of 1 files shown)


borisdj:
Thx for contrib.

`isAsync` is flag-arg pattern/hack.
Public Async and NonAsync are merged into single operation flow with protected method using arg: bool isAsync (keeps code DRY)
https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#the-flag-argument-hack

borisdj:
Version 6.1.4 published.

PR #679, opened 2021-12-06
Remove timestamptz type modification

Hoffs:
Reverts change done in https://github.com/borisdj/EFCore.BulkExtensions/commit/e70de346179c3c425e3f765b11071b02743c69cc that modified `timestamp with time zone` to `timestamp`.
This shouldnt be done as the timestamp change is intentional and was introduced in Npgsql 6 to set some constraints on how `DateTime` is used/communicated with Postgres.
See:
- https://www.npgsql.org/doc/release-notes/6.0.html#timestamp-rationalization-and-improvements
- https://www.roji.org/postgresql-dotnet-timestamp-mapping

Error in the comment is correct and should have been corrected in code rather than in the library code, as it notifies user that incorrect assumptions are made regarding date time usage. It also breaks code that uses `DateTimeOffset` for `timestamp with time zone` as Npgsql picks `TimestampHandler` for `timestamp` column field which is not supported by that handler https://github.com/npgsql/npgsql/blob/a1d366c5692cc00a5edc1ec5bc9090952c7a63e7/src/Npgsql/Internal/TypeHandlers/DateTimeHandlers/TimestampHandler.cs#L20 .

Updated tests to have a column `timestamp without time zone` that can be populated with standard `DateTime` (with kind `Local`) and another column with `DateTimeOffset` that uses `timestamp with time zone`.

Test run locally (2 failed due to: `Microsoft.Data.SqlClient.SqlException : Common Language Runtime(CLR) is not enabled on this instance` as I can't get SQL Server with CLR running):
```
Failed!  - Failed:     2, Passed:    62, Skipped:     0, Total:    64,
```

Also while debugging I found a piece of code that seems to have incorrect assumptions on `timestamp` fields which assumes that they are used for versioning/concurrency check. `timestamp` is alias for type `timestamp without time zone` in postgres, but if `HasColumnType` is set to `timestamp` populated `TableInfo` incorrectly assumes that it is used for version/concurrency check.

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
+++ b/EFCore.BulkExtensions.Tests/EFCore.BulkExtensions.Tests.csproj
@@ -10,7 +10,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer.NetTopologySuite" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="6.0.0" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="6.0.1" />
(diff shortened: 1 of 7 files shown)


borisdj:
Thx for contrib.

PR #714, opened 2022-01-09
Added global function to return shadow property values

martosource:
Added an alternative method of supplying values for shadow properties.

In my use case, I only need to set IsDelete to false for soft deletes and discriminator columns.

Using the function will save memory as not needing to add the entity to the EF change tracking.


borisdj:
Thx for contrib.
PR merged, and will be published with next NuGet till end of week.

borisdj:
Published v6.2.9

PR #792, opened 2022-04-06
fix: add value conversion to postgres adapter

### LINKED ISSUE #777: Postgres: Cannot bulk insert UInt64 values "Can't write CLR type System.UInt64 with handler type NumericHandler"
When trying to use bulk operations with postgres I cannot insert models that have unsigned longs. THis does however work normally with EFCore as it converts the value to a `numeric(20, 0)`.

When I try with bulk extensions the following exception is thrown:

> Can't write CLR type System.UInt64 with handler type NumericHandler

- EF Core: v6
- Bulk Extensions: v6.4.2
- Npgsql.EntityFrameworkCore.PostgreSQL: v6.0.3

brian-tygotech:
resolves #777 

Existing behavior has been kept (although it's not clear if this is necessary) used the same approach as in SQL Server.

Have tested locally and works. Unsure of testing strategy however, so no tests in the codebase (the tests don't run locally for me).


douglasg14b:
Just passing by, but is the meat of the change the following:

> `propertyValue = converter.ConvertToProvider.Invoke(propertyValue);`

Essentially allowing any EF value converter to be used for the value?

brian-tygotech:
Yes, that's the only functional change. This is what the SQL server adapter
does and what appears to be missing here.

PR #807, opened 2022-05-01
Enabled nullabilities, resolved 400+ warnings and error handling

ChristopherVR:
This rather large PR is to enable nullables and resolve all the warnings in the project. This PR is not related to any error per se, but rather to make it easier to determine what variables and parameters are nullable. Some errors have been suppressed in cases where I felt the value would not be null, or the warning was not valid in the context.

### DIFF
--- a/EF.BulkExtensions.IntegrationTests/EF.BulkExtensions.IntegrationTests.csproj
+++ b/EF.BulkExtensions.IntegrationTests/EF.BulkExtensions.IntegrationTests.csproj
@@ -1,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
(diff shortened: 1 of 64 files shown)


borisdj:
Nice work.
I'm reviewing it now, but you should have left namespace indentation changes for later, or for separate commit so that effective changes are separated and easier to see.

ChristopherVR:
> Nice work. I'm reviewing it now, but you should have left namespace indentation move for later, or for separate commit so that effective changes are separated and easier to see.

Do you think I should still separate the two?

borisdj:
I can remove indentation if you could then make another fork and copy other changes into it, since exiting one would be in conflict.

ChristopherVR:
> I can remove indentation if you could then make another fork and copy other changes into it, since existing one would be in conflict.

Yes that can work for me thanks!

borisdj:
I have made a new commit with switch to File-Scoped namespace, and will now close this PR.
After forking again you can split remaining changes into several commits and PRs.
For example:
First one can be current warning about XML
Then in second to enable<Nullable> and alter them.
In third remaining code cleanup and changes.
Fourth extending tests, etc...

Leave nugets version updates to me.

PR #812, opened 2022-05-08
Resolved xml warnings

ChristopherVR:
This PR is to resolve all the XML warnings by adding basic descriptions or suppressing the error where it doesn't make sense to add XML comments.

I have also enabled to setting TreatWarningsAsErrors to prevent any warnings from being checked in to the main project. Reason for this setting is because I strongly feel there should be no warnings, if there is a warning that is not valid it should be suppressed.


borisdj:
Thx for contrib.

PR #813, opened 2022-05-08
Enabled nullables

ChristopherVR:
This PR is to enable basic nullable support in the project.

1. Enabled nullable support
2. Surpressed nullable warnings in cases where we know the param/prop won't be null
3. Resolved warnings surrounding nullables
4. Threw exceptions in certain cases if the value is nullable

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
@@ -47,8 +47,8 @@ public class EFCoreBulkTest
     }
 
     [Theory]
-    [InlineData(DbServer.PostgreSQL, true)]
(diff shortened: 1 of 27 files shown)


borisdj:
I have merge first PR, but this second one, and also others, have now conflicts, not sure why.
Have you make them one after another ?

borisdj:
@ChristopherVR actually I see now that you have made separate forks/branches for each commit.
That's now practical because of possible conflict as there are now.
You should have done all on one same fork, just make changes and commits one after another.
If you can't easily resolve conflict maybe to copy code changes to main one and make them as sequential commits.

ChristopherVR:
@borisdj ,

Yeah I realised my mistake afterwards. I have resolved the conflicts on the PRs. If any additional conflicts pop up, let me know in order to resolve them.

PR #814, opened 2022-05-08
Resolved violations and applied simplifications

ChristopherVR:
This PR is to resolve violations being shown (methods needing to be marked as static, and CancellationToken should always be the lass paramater) as well as a few code simplifications.


borisdj:
Conflict. v2

ChristopherVR:
@borisdj ,

Should now be resolved

Issue #833, opened 2022-05-16
Consider caching certain reflection aspects


Danielku15:
Hi there. 

I just integrated EFCore.BulkExtensions in the hope to improve the performance in some heavyload operations we are having. As part of the profiling I did, I noticed that on each individual `BatchUpdate` call, will create a new `BatchUpdateCreateBodyData` and subsequently `TableInfo` and `FastProperty`.

Let's say we update 2000 times objects in a loop using a statement like `dbContext.MyObjects.Where(...).BatchUpdate(x => new MyObject { Prop = someValue })`. This will result in:

* 2000 `BatchUpdateCreateBodyData` instances
* 2000 `TableInfo` instances
* 2000 `FastProperty`
* 4000 newly compiled delegates for the getter/setter in the `FastProperty`

This is quite a waste of resources (memory and CPU/time wise) and also the GC will not be happy. It would be quite a big task to have an automatic query caching/compilation like EF does internally, but maybe there could be some low-hanging fruits to improve:

`FastProperty` can be cached globally. This would already reduce the 2000 `FastProperty` and 4000 compiled lambdas into 1 `FastProperty` and 2 lambdas. A static `ConcurrentDictionary<PropertyInfo, FastProperty>` with a `GetOrAdd` instead of a direct constructor call should do the job. 

An API like the old `CompiledQuery`[^1] or `EF.CompileQuery`[^2] could be added to EFCore.BulkExtensions to reduce also the `TableInfo` and `BatchUpdateCreateBodyData` instances. Code paths which are executed in high frequency could be benefit a lot from such a feature.  

I can provide some more detailed figures if needed, but already somebasics show that there is a big performance hit in this basic case where we update hundreds of records in a loop using `BatchUpdate`:
![image](https://user-images.githubusercontent.com/674916/168621941-f072bfa8-3d62-4536-ba74-17573e47e945.png)

Let me know what you think about this topic and if there would be interest in adding such improvements.

[^1]: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/ef/language-reference/compiled-queries-linq-to-entities
[^2]: https://gunnarpeipman.com/ef-core-compiled-queries/

borisdj:
Using Dicts for loop is always a good idea, will take a look into it.

borisdj:
PR merged.

PR #860, opened 2022-06-23
Global cache for FastProperty instances.

Danielku15:
This PR addresses some basics of #833 by caching `FastProperty` instances globally in a `ConcurrentDictionary`. 

I did not invest any efforts yet to allow pre-compilation of certain operations to also reduce the creation of other objects like the `TableInfo`


borisdj:
Thx for contrib.

PR #870, opened 2022-06-30
Fix PropertiesToExcludeOnUpdate column selection issue with postgres.

alexanderlunden:
`tableInfo.PropertyColumnNamesUpdateDict` maps model/POCO property names (key) to database column names (value), and `GetColumnList()` returns database column names, thus we need to match the dictionary entries by value and not key, when selecting which columns should have their values updated.


borisdj:
Thx for contrib.

soligto:
Hello. I faced this bug recently and glad to find out it's already fixed. Is it possible that a fixed version will be released any time soon?

PR #872, opened 2022-07-01
Add support for custom type identity properties (using converters)

janno-p:
When using custom types as identity values (for example struct wrapping an integer value), bulk operations fail because extension is trying to cast custom type to the `long` type instead. Since ValueConverters are available and used for regular properties, added some code to use them when dealing with identity values aswell.

Same issue and fix with shadow foreign key properties in sql server adapter.


borisdj:
Thx for contrib

PR #881, opened 2022-07-14
Only bulkInsert support for mySql

saeidbabaei-dev:
I have done BulkInsert method for mysql with out graph support.
please review my changes. if you agree with me, i will continue to support other bulk operations.
thank you in advance.


borisdj:
Nice work.

genaray:
Can you support SaveChanges for MySQL ? 
This is what i use primarly due to its easy useage ^^ Would be great

saeidbabaei-dev:
Yes, i am going to support SaveChanges method for MySQL as soon as possible.

EV0755:
Will you support  BatchUpdate for MySQL?

borisdj:
BatchUpdate method will soon, in a few months as of EF7, be supported directly in EF Core, so no need for it here.
Probably both(Update/Delete) Batch ops will become deprecated as an Extension.

PR #889, opened 2022-07-19
Add graph option to BulkInsert method

saeidbabaei-dev:
please review my code and give me feedback.


borisdj:
Checked and merged.
Also I've added a Test for MySql, for now only for BulkInsert.
Segment for BulkInsertOrUpdate is currently commented since if throws exception:
`Table 'efcorebulktest.itemtemp5ebc6d78' doesn't exist`.

Issue #770, opened 2022-03-06
Unable to BulkInsertAsync with a basic Owned Entity model - Postgres


miguellira:
I apologize up front if this turns out to be just a basic newbie question. However, I can't seem to configure `BulkInsertAsync` to work with a model that has a simple `Owned` entity.

The following repro has a model containing a single `DbSet` with an `Owned` entity. No additional configuration. I've tried all sorts of `BulkConfig` settings but can't manage to get this to work.

**Program.cs**
```c#
using EFCore.BulkExtensions;
using Microsoft.EntityFrameworkCore;

using var dbContext = new SchoolDbContext();

await dbContext.Database.EnsureCreatedAsync();

var students = new[] {
    new Student { Name = "Jack", Address = new Address { Line1 = "125 UpTheHill St." }},
    new Student { Name = "Jill", Address = new Address { Line1 = "125 UpTheHill St.", Line2 = "Floor 250" }}
};

var bulkConfig = new BulkConfig
{
    SetOutputIdentity = true,
    PropertiesToInclude = new List<string> {
        $"{nameof(Address)}.{nameof(Address.Line1)}",
        $"{nameof(Address)}.{nameof(Address.Line2)}"
    }
};

await dbContext.BulkInsertAsync(students, bulkConfig);

public class SchoolDbContext : DbContext
{
    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
    {
        optionsBuilder.UseNpgsql("Host=localhost;Port=5432;Database=school;Username=postgres;Password=postgres;Include Error Detail=true;Log Parameters=true;Application Name=EFCore.BulkExtensions-Tester")
                      .EnableDetailedErrors(true)
                      .EnableSensitiveDataLogging(true)
                      .LogTo(Console.WriteLine);
    }

    public DbSet<Student> Students => Set<Student>();
}

public class Student
{
    public int Id { get; set; }
    public string Name { get; set; } = null!;
    public Address Address { get; set; } = null!;
}

[Owned]
public class Address
{
    public string Line1 { get; set; } = null!;
    public string? Line2 { get; set; }
}
```
**Students.csproj**
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

[...]

Issue #695, opened 2021-12-17
Unable to bulk insert entities with collection of owned entities


fdipuma:
It seems I'm unable to do BulkInserts / BulkSaveChanges with EFCore.BulkExtensions 6.2.2 (Sql Server provider) and collections of Owned entities.

This is caused by multiple factors:
1. I receive a `NullReferenceException` when using `BulkSaveChanges`:
```
at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlOperationsServerAdapter.InnerGetDataTable[T](DbContext context, Type& type, IList`1 entities, TableInfo tableInfo) in D:\Dev\Github\EFCore.BulkExtensions\EFCore.BulkExtensions\SQLAdapters\SQLServer\SqlServerAdapter.cs:line 541
   at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlOperationsServerAdapter.GetDataTable[T](DbContext context, Type type, IList`1 entities, SqlBulkCopy sqlBulkCopy, TableInfo tableInfo) in D:\Dev\Github\EFCore.BulkExtensions\EFCore.BulkExtensions\SQLAdapters\SQLServer\SqlServerAdapter.cs:line 418
   at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlOperationsServerAdapter.InsertAsync[T](DbContext context, Type type, IList`1 entities, TableInfo tableInfo, Action`1 progress, CancellationToken cancellationToken, Boolean isAsync)
   at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlOperationsServerAdapter.InsertAsync[T](DbContext context, Type type, IList`1 entities, TableInfo tableInfo, Action`1 progress, CancellationToken cancellationToken, Boolean isAsync) in D:\Dev\Github\EFCore.BulkExtensions\EFCore.BulkExtensions\SQLAdapters\SQLServer\SqlServerAdapter.cs:line 104
   at EFCore.BulkExtensions.SQLAdapters.SQLServer.SqlOperationsServerAdapter.InsertAsync[T](DbContext context, Type type, IList`1 entities, TableInfo tableInfo, Action`1 progress, CancellationToken cancellationToken) in D:\Dev\Github\EFCore.BulkExtensions\EFCore.BulkExtensions\SQLAdapters\SQLServer\SqlServerAdapter.cs:line 31
[...]

Issue #566, opened 2021-06-02
Unable to use bulk read with owned type


sandeepk87:
Hi,
I am trying to use bulk read filter by owned type value in MSSQL.

  `
        
       public class User : Entity
       {
        protected User (){}
        public User(string name, EmailAddress email)
        {
            Name = name;     
            Email = email;
        }
        public void Update(string name, EmailAddress email)
        {
          Name = name;
          Email = email;
        }

        public string Name { get; private set; }
        public  EmailAddress Email { get;  set; }
       }

        [Owned]
       public class EmailAddress :ValueObject
       {
        public string Value { get; }

        public EmailAddress(string value)
        {
            if (EmailValidator.IsValid(value) == false)
                throw new  EmailAddressIsNotValidException( value);
            Value = !string.IsNullOrWhiteSpace(value) ? value.ToLowerInvariant() : value;
        }

        protected EmailAddress()
        {

        }
        public override string ToString()
        {
            return Value;
        }

        protected override IEnumerable<object> GetEqualityComponents()
        {
            yield return Value;
        }
        }

   `
  
  I am trying to query User Table filter by email list - 
  
    ` var entities = emailList.Select(f => new User(null, new EmailAddress(f))).ToList();
      var bulkConfig = new BulkConfig() { UpdateByProperties = new List<string>() { nameof(User.Email) } };
      await _context.BulkReadAsync(entities, bulkConfig);
   `
 
  I am getting error - "**Sequence contains no matching element**"
  
  Is there any other config to be set while using owned type? 

  Library Version -
   "EFCore.BulkExtensions" Version="5.1.8" 
   "Microsoft.EntityFrameworkCore" Version="5.0.5"

borisdj:
Not supported currently.

borisdj:
Should be fixed with v6.5.5.

PR #893, opened 2022-07-20
Make owned types (in same table) work in PostgreSQL.

### LINKED ISSUE #795: PostgreSQL BulkInsertOrUpdateAsync on entity with a default value causes InvalidOperationException
Given the following code:

**Entity**
```csharp
public class ExampleEntity
{
    public Guid Id { get; set; }
    public int Field { get; set; }
    public bool Other { get; set; }
}
```

**Entity fluent configuration**
```csharp
builder.Property(x => x.UserId).ValueGeneratedNever();
builder.Property(x => x.Field).IsRequired();
builder.Property(x => x.Other).IsRequired().HasDefaultValue(false);
```

**Example**
```csharp
await context.BulkInsertOrUpdateAsync(new List<ExampleEntity>
{
    new ExampleEntity
    {
        Id = Guid.NewGuid(),
        Field = 1,
        Other = False
    },
    new ExampleEntity
    {
        Id = Guid.NewGuid(),
        Field = 1,
        Other = False
    }
});
```

The exception occurs:

```
System.InvalidOperationException: The binary import operation was started with X column(s), but X value(s) were provided.
```

This happens because fields with a default value in the database are excluded from the insert statement https://github.com/borisdj/EFCore.BulkExtensions/blob/abe6ca5ed73634bca19f47f8335a736d4ca5d119/EFCore.BulkExtensions/SQLAdapters/PostgreSql/PostgreSqlAdapter.cs#L68

rwasef1830:
Make owned types (in same table) work in PostgreSQL.
Fix identity annotation identification. Fixes #795 and #695 and #770 and #566 and regression from #772. With passing unit test.

@borisdj

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTestAtypical.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTestAtypical.cs
@@ -417,6 +417,7 @@ public class EFCoreBulkTestAtypical
     [Theory]
     [InlineData(DbServer.SQLServer)]
     [InlineData(DbServer.SQLite)]
+    [InlineData(DbServer.PostgreSQL)]
     private void OwnedTypesTest(DbServer dbServer)
     {
         ContextUtil.DbServer = dbServer;
@@ -426,6 +427,10 @@ public class EFCoreBulkTestAtypical
         {
             context.Truncate<ChangeLog>();
             context.Database.ExecuteSqlRaw("TRUNCATE TABLE [" + nameof(ChangeLog) + "]");
+        }
+        else if (dbServer == DbServer.PostgreSQL)
+        {
+            context.Truncate<ChangeLog>();
         }
         else
         {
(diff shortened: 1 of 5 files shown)


rwasef1830:
I also discovered postgres integration doesn't properly implement "with graph" :-(

borisdj:
Thx for PR, Great work.
Merged and v6.5.5 published. 

PS
Yes EF generated Sql is not always as expected, or as would a person write it directly.
Had similar issue before.
You can do first BulkInsert, then BulkRead.

borisdj:
There was a bug, regarding Identity column, introduced a few days ago with PR for MySql support.
I only found it this morning, it's now fixed so take the sync with latest source.

MWRouin:
Hello!

This did not work for me, i am trying to bulk insert  entities that has some owned properties (1 to 1 in same table) but they are being ignored and stored as null!

PR #909, opened 2022-07-28
Owned types with spatial data

TheJoeIaut:
Hi,

I found, that Owened Types that contain spatial data like

```
[Owned]
public class TrackerLocation
{
    public string LocationName { get; set; } = null!;
    public Point Location { get; set; } = null!;
}
```
are failing when used in types that are Bulk Inserted/Updated.

The problem was, that the Geometry Logic, that is used in normal Properties, was not used on Properties of Owned Types.

I added that logic and created a Unit Test for it.

It is also only for SQL Server (since the Geometry Handling for normal Properties is only working in SQL Server)


borisdj:
Thx for contrib.

PR #913, opened 2022-08-06
will support also BulkUpdate, BulkDelete and BulkInsertOrUpdate methods for mysql

saeidbabaei-dev:
adding bulk update and bulk delete methods


borisdj:
All basic CRUD seem to work fine.
One situation that did not work is for _Update_ or _InsertOrUpdare_ with custom **UpdateByProp** (no Unique Index)
[SO_mysql-insert-on-duplicate-update-for-non-primary-key](https://stackoverflow.com/questions/33495194/mysql-insert-on-duplicate-update-for-non-primary-key)

I've supported it by temporary adding UNIQUE CONSTRAINT, similarly is done for PG.
Added Segment of MySql Test with 'entities4'.

saeidbabaei-dev:
Thank you.
If you dont mind, i wanna have a conversation with you.

borisdj:
Sure, you can text me on djurdjevicboris@hotmail.com and we can arrange a time.

PR #943, opened 2022-09-07
Add missing .ConfigureAwait(false) everywhere

0xced:
Also set the CA2007 severity as warning which will end up as an error since `TreatWarningsAsErrors` is enabled. (But disable them in the test project where they are not required.)

See also [Overview of .NET source code analysis][1] for more information about the .NET analyzers and how to configure them.

[1]: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,4 @@ end_of_line = lf
 trim_trailing_whitespace = false
 insert_final_newline = true
 csharp_style_namespace_declarations = file_scoped:warning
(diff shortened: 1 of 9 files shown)


borisdj:
Could you resolve the conflict, or I can add the changes manually since I'm currently doing some refactoring.

0xced:
I just rebased this pull request on master, resolved the conflicts and force pushed so that it can be merged.

PR #957, opened 2022-09-23
[Postgresql] Remove FROM in batch update if no JOINs to other tables 

achelminski:
In postgresql, if batch update has no joins to other table, 'target table' should not be repeated.

According to postgresql documentation:
 "Do not repeat the target table as a from_item unless you intend a self-join"

 https://www.postgresql.org/docs/current/sql-update.html


borisdj:
Can you resolve the conflicts.

borisdj:
Since Batch ops are becoming deperacated as an extension, because they are natively supported as of EF7 this issue no longer very relevant, but still I have added the change into repo, commit note: *PG removed FROM in Batch, PR 957*

PR #982, opened 2022-10-20
Gracefully Handle a wrapped QueryProvider (Delegate Decompiler support during Batch operations)

cmeyertons:
The [DelegateDecompiler](https://github.com/hazzik/DelegateDecompiler) library ends up wrapping the EntityQueryProvider in its own instance.  

The reflection behavior of looking up the private `_queryCompiler` field fails in this case. 

This PR aims to add graceful handling of a wrapped EntityQueryProvider instance for this use case.

DelegateDecompiler does the same pattern for all of its flavors of EF (EF6, EFCore, EFCore5+) so testing against DelegateDecompiler.EFCore5 is probably sufficient in this scenario.

### DIFF
--- a/EFCore.BulkExtensions.Tests/DelegateDecompilerTests.cs
+++ b/EFCore.BulkExtensions.Tests/DelegateDecompilerTests.cs
@@ -1,0 +1,20 @@
+﻿using DelegateDecompiler.EntityFrameworkCore;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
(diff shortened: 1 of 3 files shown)


borisdj:
As of EF 7 there is now ExecuteUpdate and ExecuteDelete so Batch-Update/Delete will probably become Deprecated.

borisdj:
Still the fix is added into repo, commit note: *Batch DelegateDecompiler support, PR 982*
Thx for contrib.

cmeyertons:
@borisdj thank you!

Issue #1004, opened 2022-11-15
TimeStamp on tables without indentity


VidVrbnjak:
If your table does not have a identity column, in my case we use uniqueidentifiers as the PK, then the TimeStamp object on the BulkConfig does not get populated, because TableInfo.LoadOutputData checks if your table has a idententity column and only then loads the data.
I don't think it should be required that you have a Identity column in your table if you want to use TimeStamps

PR #1005, opened 2022-11-16
TimeStamp on tables without indentity #1004

VidVrbnjak:
Expanded the condition in TableInfo to load data even when we don't have a identity column, but still want TimeStamps to be handled


borisdj:
That config is not quite suitable for this use case.
DoNotUpdateIfTimeStampChanged when set means not to apply chaneg to row in Db if TimeStamp was changed since the read.
Updated list from DB is another issue, maybe to add explicit property in config.

Issue #993, opened 2022-11-09
net7 support


sommmen:
Hiya,

net7 is out so i though i'd make an issue to track net7 and ef core 7 support.

Webreaper:
Bulkextensions already supports .Net 7. Unfortunately, though, it seems to be explicitly pulling in dependencies for RC2, which is causing a package downgrade compile failure when using the full .Net 7 GA. :(

mkalinski93:
Net7 will be very nice.

@Webreaper I´ve tried implementing it, but as you´ve stated correctly the dependency downgrade is the main issue why it´s not working on my end. I think it was the Microsoft.EntityFrameworkCore.Relational library which causes a lot of trouble. I will fork it and try to overcome this issue

bdbvb:
I'm getting the following build warnings after updating an existing project to .NET 7. The mentioned dependency is used by EFCore.BulkExtensions.

`Detected package version outside of dependency constraint: Pomelo.EntityFrameworkCore.MySql 6.0.2 requires Microsoft.EntityFrameworkCore.Relational (>= 6.0.7 && < 7.0.0) but version Microsoft.EntityFrameworkCore.Relational 7.0.0 was resolved.
`

mikebrownNCDOT:
BulkExtensions can't be updated for net7 support until the Pomelo MySql library gets theirs working. If you want to follow that (should be very soon), go here: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1717

Webreaper:
I'm building fine now. Can't remember what I changed but it all works.

Syzuna:
I've just got rid of the MySQL implementation as I dont need it anyway and I am using my own build for now. 
besides the MySQL driver issue everything else was smooth sailing

jr01:
and if you update to .NET 7 and also target EF Core 7 and use BulkUpdateAsync, BulkInsertOrUpdate, or BulkInsertOrUpdateOrDelete you'll get a runtime exception:
```
System.MissingMethodException: Method not found: 'Microsoft.EntityFrameworkCore.Storage.IRelationalCommand Microsoft.EntityFrameworkCore.Query.QuerySqlGenerator.GetCommand(Microsoft.EntityFrameworkCore.Query.SqlExpressions.SelectExpression)'.
  Stack Trace: 
   IQueryableExtensions.ToParametrizedSql(IQueryable query)
[...]

PR #1018, opened 2022-12-09
Feature/net7

sommmen:
Should go away at solving #993 

I set the TFM to both net6 and net7 and then moved all packages so that they'd use the right nuget version.

I ran the tests for SqlServer, MySql and SQLite, these were all good except a couple failed for net7 (net6 were all green).
I skipped Postgres tests because i don't have a db for that running.

image

I had some SQLite issues `The process cannot access the file because it is being used by another process.` - i did not look into this and skipped those tests.
These seemed to be easy to fix, like for example slightly different SQL being generated in net7:

image

I then installed this package into my own net7 project, it uses a lot of different configurations of BulkOptions and I did not see any issues.

I named this version 7.0.0-beta.1, the same name as the `Pomelo.EntityFrameworkCore.MySql` package. I'd be great if at some point we'd have atleast a beta package that supports .net7.

Feel free to do anything with this PR as you see fit.

### DIFF
--- a/EFCore.BulkExtensions.Common/EFCore.BulkExtensions.Common.csproj
+++ b/EFCore.BulkExtensions.Common/EFCore.BulkExtensions.Common.csproj
@@ -29,13 +29,13 @@
   </PropertyGroup>
   
   <ItemGroup>
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
-	<PackageReference Include="NetTopologySuite.IO.SqlServerBytes" Version="2.1.0" />
-    <PackageReference Include="EntityFrameworkCore.SqlServer.HierarchyId" Version="3.0.1" />
+      <PackageReference Include="NetTopologySuite.IO.SqlServerBytes" Version="2.1.0" />
   </ItemGroup>
 
-    <!-- net 6.0 -->
-    <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+  <!-- net 6.0 -->
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
+        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="6.0.0" />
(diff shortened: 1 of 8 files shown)


Manny651:
EF Core 7 only requires .NET 6. It should not depend on the target framework.

benmccallum:
@Manny651 , yes, but this library relies on EF Core 7 and there must be a binary breaking change.

benmccallum:
Ah sorry, I think I might've misunderstood what you're saying. You're saying we shouldn't need to associate the EF Core version being used with a particular target .NET version...

benmccallum:
Honestly, I think this PR, given it's a major version bump to v7, should just target net7.0 and EF Core 7 only. It gets messy trying to support other configurations. The migration from net6.0 to net7.0 is easy enough and worst-case features can be ported back to a branch off 6.x if needed in future.

Manny651:
> That seems to be the approach @borisdj has taken thus far looking at the compat table in the README

In the compat table, The "For projects targeting" explains the normal usage, but the target framework has always been the lowest TFM required by EF Core. ie the 5.x version targeted NetStandard 2.1.

sommmen:
net7 is STS and net6/7 will both be supported till 2024 so it would be nice to support both of them.
Reading through this it seems staying on net6 (given its compatible with net7) and updating only EF to EF7 would be the way to go.

Right now running there is a missing method exception when using EF7 because of this;
image

i'll take a look and see if just updating to EF7 would work when i have time. Also maybe i can cast to to Expression and it'll find the right method.

AraHaan:
Also, I use EFCore 7 with .NET 6 (although if they do not have .NET 6's runtime installed it will roll forward to .NET 7's runtime). As such supporting both even for me is a must.

sommmen:
> Also, I use EFCore 7 with .NET 6 (although if they do not have .NET 6's runtime installed it will roll forward to .NET 7's runtime). As such supporting both even for me is a must.

Yes, got an alternative PR #1044

PR #1044, opened 2022-12-23
fix: find EF7 GetCommand via reflection as workaround

sommmen:
Should help with #993  

As discussed in #1018 when using EF7 the signature of `QuerySqlGenerator.GetCommand` has changed. 
This prevents using EF7. 

image

This PR introduces a helper class that finds the EF7 `GetCommand` via reflection and caches it as a delegate. This will then be used instead of a direct call to `GetCommand`. Under EF6 the direct call will be used as it always has.

I've tested this with NET6 EF6 and NET7 EF7. 
The code is simple and should not pose great risk.

This does impose a small performance hit, regarding the nullcheck and calling a delegate in EF7,
but these are so small it outweighs the benefit.


sommmen:
CC @Manny651 @benmccallum @borisdj

Issue #1085, opened 2023-02-01
SqliteParameter Issue When Object Parameter Uses HasColumnName After 6.6.5


corykroll:
I've been having an issue with any version after 6.6.5.

I have an object in which a parameter is named differently in the POCO than in the database.
Let's say I define my database object with a nullable enum parameter as follows:

```
[Table("PassingRecord")]
public partial class PassingRecord
{
     public PassingType? PassingType { get; set; }
}
```

Then, in the DbContext, note that this is named differently in the database, since I store the enum value as an integer:

```
modelBuilder.Entity<PassingRecord>()
.Property(x => x.PassingType).HasColumnName("PassingTypeID");
```

This used to work.  But after upgrading past 6.6.5 I instead get the following error instead when doing bulk inserts with this model:

`System.IndexOutOfRangeException: 'A SqliteParameter with ParameterName '@PassingType' is not contained by this SqliteParameterCollection.'`

It is as if it doesn't recognize the property mapping anymore.

I am targeting NET 7.
If I use the 6.6.5 NuGet it works again, even targeting NET 7.

mtelle:
Hi, I have found that problem yesterday, i jave created a PR for that fix 
#1086

Naveen61097:
Hi Team, Any updates on this? It would be better if it is fixed ASAP.

borisdj:
PR merged, thx for contrib.
New nuget will be released in a few days.

Btw sql params are created using Property names (@SomeName), but I figure they could be created using Column names.
What would you consider to be better convention and there are special usecases.

borisdj:
v7.0.3 is published

PR #1086, opened 2023-02-02
Fix Sql Lite command when property of a column is camelcase but the mapping is uppercase

mtelle:
Hi,
I have found a possible bug for SqlLite. In my case my model of the entities of EF are camelcase, but the mapping is upercase.

Example:
    public partial class Actor
    {
        public int Id { get; set; }
    }
Mapping
        public override void Configure(EntityTypeBuilder<Actor> builder)
        {
            builder.ToTable("ACTORS", schema);

            builder.HasKey(p => new { p.Id });

            // Properties:

            builder.Property(p => p.Id)
                     .HasColumnName("ID")
                     .IsRequired()
                     .ValueGeneratedOnAdd();
}

With this configuration there is a mistmatch with the bulkcommands query and the parameters:
image

The command query is like this:
UPDATE [ACTORS] SET [ID] = @Id  WHERE [ID] = @Id;

The command parameters are Uppercase, but must be the original name of the property Id instead of ID.

With this simple change now it create the good command and the good paratemer names.

### DIFF
--- a/EFCore.BulkExtensions/SqlAdapters/SQLite/SqLiteAdapter.cs
+++ b/EFCore.BulkExtensions/SqlAdapters/SQLite/SqLiteAdapter.cs
@@ -377,7 +377,7 @@ public class SqliteOperationsAdapter : ISqlOperationsAdapter
 
             if (propertyEntityType != null)
             {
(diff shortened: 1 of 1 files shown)


mtelle:
#1085

Issue #1014, opened 2022-12-07
postgre, BulkUpdate inserts records that are not in the table


sandiks:
can you tell me if this is the correct behavior , when using this method, rows are inserted that are not in the table

geis01:
Hello,

This seems consistent with my experience from #1006

borisdj:
Can you write reproducible example or test.

borisdj:
Actually should be fixed now with v7.0.3

PR #1091, opened 2023-02-14
Fix Postgres BulkUpdate

### LINKED ISSUE #1006: Postgresql: PropertiesToInclude doesn't work with BulkUpdate
I'm using BulkUpdate with a BulkConfig object containing a PropertiesToInclude limiting my update to a single ("total") field. But it seems  the generated sql will try to update other fields anyway:

```
23502: null value in column "status" of relation "order" violates not-null constraint

DETAIL: Failing row contains (19853, null, 0, null, 1, null, 0, null, null).
```

EDIT: 

I believe it is related to postgres attempting an insert with a row containing only my two fields (the key and the field I wish to update) - as the not null constraint fails, it seems it never gets to do the "ON CONFLICT .. DO UPDATE" part of the generated sql. If I change the MergeTable method in the SqlQueryBuilderPostgreSql class with a specific OperationType.Update handling, I can get it to work with an update statement. Something like the following:

```
var columnsListEquals = GetColumnList(tableInfo, OperationType.Insert);
var columnsToUpdate = columnsListEquals.Where(c => tableInfo.PropertyColumnNamesUpdateDict.ContainsValue(c)).ToList();

if (operationType == OperationType.Update)
{
	string updateColumns = string.Join(",", columnsToUpdate.Select(col => 
		$"{col.Replace("[","\"").Replace("]","\"")}={tableInfo.FullTempTableName.Replace("[",@"""").Replace("]",@"""")}.\"{col.Replace("[",@"""").Replace("]",@"""")}\"")
	);
	string whereCondition = string.Join(" AND ",
		tableInfo.PrimaryKeysPropertyColumnNameDict.Values.ToList().Select(col => 
			$"{tableInfo.FullTableName.Replace("[",@"""").Replace("]",@"""")}.\"{col.Replace("[",@"""").Replace("]",@"""")}\" = " +
			$"{tableInfo.FullTempTableName.Replace("[",@"""").Replace("]",@"""")}.\"{col.Replace("[",@"""").Replace("]",@"""")}\"")
	);
	q =  $"UPDATE {tableInfo.FullTableName.Replace("[",@"""").Replace("]",@"""")} SET {updateColumns} " +
		 $"FROM {tableInfo.FullTempTableName.Replace("[",@"""").Replace("]",@"""")} {(!string.IsNullOrEmpty(whereCondition) ? "WHERE " : "")}{whereCondition}";
}
```

tgardner:
Fixes #1006 
BulkUpdate fails in postgres with not-null constraints for excluded columns

### DIFF
--- a/EFCore.BulkExtensions.Tests/SqlQueryBuilderPostgreSqlTests.cs
+++ b/EFCore.BulkExtensions.Tests/SqlQueryBuilderPostgreSqlTests.cs
@@ -51,6 +51,20 @@ public class SqlQueryBuilderPostgreSqlTests
             expected = expected.Replace(" LIMIT 1", "");
         Assert.Equal(expected, actual);
     }
+
+    [Fact]
+    public void MergeTableUpdateOnlyTest()
+    {
+        TableInfo tableInfo = GetTestTableInfo();
+        tableInfo.IdentityColumnName = "ItemId";
+        string actual = SqlQueryBuilderPostgreSql.MergeTable<Item>(tableInfo, OperationType.Update);
+
+        string expected = @"UPDATE ""dbo"".""Item"" SET ""Name"" = ""dbo"".""ItemTemp1234"".""Name"" " +
+                          @"FROM ""dbo"".""ItemTemp1234"" " +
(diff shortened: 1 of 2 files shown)


tgardner:
In addition, should also resolve #1014

PR #1142, opened 2023-04-23
Use Postgresql Temp table and fix not drop temporary table when UseTe…

gortha:
Fix when UseTempDB is true the temp table created is not dropped and it is not a Postgresql Temporary table


borisdj:
Thx for contrib.
New Nuget will be published in a few days.

PR #1156, opened 2023-05-24
Get Postgresql default schema from Search Path

cime:
Fix for Postgresql schemas, when schema is set in a Search Path parameter of a connection string.
Without this the Search Path is ignored (we set it at runtime - multitenant app) and defaults to "public", which in our case is wrong.

https://www.npgsql.org/doc/connection-string-parameters.html#misc

Examples:

| Connection String | Default Schema |
| --- | --- |
| `Host=myserver;Username=mylogin;Password=mypass;Database=mydatabase;SearchPath=myschema` | `myschema` |
| `Host=myserver;Username=mylogin;Password=mypass;Database=mydatabase;SearchPath=myschema,public` | `myschema` |
| `Host=myserver;Username=mylogin;Password=mypass;Database=mydatabase` | `public` |

### DIFF
--- a/EFCore.BulkExtensions/TableInfo.cs
+++ b/EFCore.BulkExtensions/TableInfo.cs
@@ -14,6 +14,7 @@ using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Npgsql;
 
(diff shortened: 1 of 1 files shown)


borisdj:
Thx for contrib.

PR #1295, opened 2023-10-17
Fix SetSynchronizeSoftDelete option

kvpt:
When using SetSynchronizeSoftDelete option the query fail with the following message.
> Incorrect syntax near ']'

After dumping the sql and locating the place in the code where the problematic statement was build, it appear that there is an hardcoded 'i' value instead of a table alias.
https://github.com/borisdj/EFCore.BulkExtensions/blob/b622a0c1fac2002a8cc62b9c7847762d0871e901/EFCore.BulkExtensions/SqlAdapters/SqlQueryBuilder.cs#L403

So if the generated alias of the entity is 'i' (like in the current unit test) it's works.
But if not the replace fail and an invalid sql statement is generated.

This fix replace the hardcoded value with the correct table alias for the update statement used when the SetSynchronizeSoftDelete is defined fixing the issue.

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTest.cs
@@ -694,6 +694,18 @@ public class EFCoreBulkTest
         var list = context.Items.Take(2).ToList();
         Assert.True(list[0].Quantity != 0);
         Assert.True(list[1].Quantity == 0);
+
+        // TEST Alias
+        context.Entries.Add(new Entry { Name = "Entry_InsertOrUpdateOrDelete" });
+        context.SaveChanges();
+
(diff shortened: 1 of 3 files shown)


borisdj:
Thx for contrib.

kachalkov:
hi guys, awesome library, how do we get version with this fix?

borisdj:
New nuget will be released in a few days.

kachalkov:
Awesome, thank you so much.

kachalkov:
Hi @borisdj, just wondering how I can track this fix in the latest package?

kachalkov:
> New nuget will be released in a few days.

@borisdj 
I can see latest package released, but it is targeting .net 8 only, are there v7 in the pipeline or we need migrate to 8?

PR #1315, opened 2023-11-14
Ef core6.0 mysql create table

YangLius:
Update MySqlQueryBuilder.CreateTableCopy
Because "Create Table NewTableName Select * FROM OldTableName LIMIT 0;" MySql RDS is not supported in aliyun, so change it to "Create Table NewTableName LIKE OldTableName;"


borisdj:
Hi, thx for contrib.
Could you make new PR without changes of Spacing if files so that real Change is easily visible as only Diff.

borisdj:
Added LIKE syntax

PR #1317, opened 2023-11-16
Fix soft delete feature on linux by using cross-platform NewLine

kvpt:
Found another bug in the soft delete feature.
The newline \r\n was hardcoded in the code.
It is problematic as it's platform dependent so it currently only works on Windows.
Replaced it with Environment.Newline to fix that.


borisdj:
Thx for contrib.

Issue #899, opened 2022-07-22
BulkSaveChanges errors not following FK dependency order


bocca:
When saving changes with inserts and/or deletes of two or more entity types joined by foreign keys, BulkSaveChanges can lead to referential integrity errors like this:
```
The MERGE statement conflicted with the FOREIGN KEY constraint "FK_XXX"
```
Example:
```csharp
dbContext.Add(new A { B = new B() })
dbContext.BulkSaveChanges()
```
That code runs fine if `Entries()` returns `{B, A}`, but throws error if `Entries()` returns `{A, B}`.

I have inspected source code of `DbContextBulkTransactionSaveChanges`, and there is nothing to ensure the correct order. The order follows the order of first appearance of each entity type in `DbContext.ChangeTracker.Entries()`. But, as far as I know, `Entries()` does not ensure any particular order.

The solution is to sort the inserts/deletes following FK dependency order. The complexity of the solution required varies with different cases:
1. Self-referencing loops in same entity type (for example parent/child node of trees) can be executed in one bulk statement without error. No problem.
2. Schemas with no FK loops between the entity types involved (besides self-referencing loops), ca be resolved, at most, with 3N bulk operations (one insert, one update and one delete for each entity type). We only have to sort the 3N operations using topological sorting of a graph of entities types and FKs from the DbContext Model. If the algorithm detects loops, an exception could be thrown.
3. Even with FK loops, theoretically, can be resolved too, but requires to make a graph, not of schema, but of individual objects, spliting inserts and deletes in several bulk operation. This is very tricky, and completely anihilates the advantages of using bulk operations.

I think at least we can achieve (1) and (2).

bocca:
Simplified workaround code solution for (1) and (2), using nuget package MedallionTopologicalSort.
```csharp
private static Type GetNonProxyType(Type type) => type.Namespace == "Castle.Proxies" ? type.BaseType! : type;

public static int BulkSaveChanges(DbContext dbContext)
{
    // Group entities to save by bulk operation (entity type and state)
    var states = new[] { EntityState.Added, EntityState.Modified, EntityState.Deleted };
    var groups = dbContext.ChangeTracker.Entries()
        .Where(x => states.Contains(x.State))
        .GroupBy(a => new { EntityType = GetNonProxyType(a.Entity.GetType()), a.State },
            (entry, group) => new
            {
                entry.State,
                Entities = group.Select(a => a.Entity).ToList(),
[...]

PR #1345, opened 2023-12-08
Proposed fix to SaveChangesAsync entity ordering for deletions

filmstarr:
Reverse deletion order of outer enumerable of entity groups rather than inner lists of entities when invoking BulkSaveChangesAsync.

#899


borisdj:
PR merged, thx for contrib

filmstarr:
🙌 Awesome, and thank you @borisdj for an excellent library! 👏

Issue #1337, opened 2023-12-05
ExecuteWithGraph: InvalidOperationException with model with subtype within owned type


fretje:
I've tried to make a minimal reproducible sample. The crux of the issue (I think) is that my model has a subtype (an `ICollection<Email>` within an owned type (`ContactMethods`), and somehow the foreign keys (`ContactMethodsCustomerId`) of the `Email` entities don't get filled before sending them to sqlbulkcopy?

```c#
using EFCore.BulkExtensions;
using Microsoft.EntityFrameworkCore;

var connectionString = "Server=localhost;Database=BulkExtensionsTestDb;Trusted_Connection=True;TrustServerCertificate=True;MultipleActiveResultSets=true;";
using var context = new MyDbContext(new DbContextOptionsBuilder().UseSqlServer(connectionString).Options);

context.Database.EnsureCreated();

var bulkConfig = new BulkConfig { IncludeGraph = true, CalculateStats = true };
Customer[] entities = [
    new()
    {
        CustomerId = "test1",
        ContactMethods = new()
        {
            HomePhone = "homephone1",
            EmailAdresses = [
                new Email() { Address = "email1", Type = "emailtype1" },
                new Email() { Address = "email2", Type = "emailtype2" },
            ]
        },
    }
];

await context.BulkInsertOrUpdateAsync(entities, bulkConfig);


public class MyDbContext(DbContextOptions opts) : DbContext(opts)
{
    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Customer>(c => 
            c.OwnsOne(x => x.ContactMethods, 
                cm => cm.OwnsMany(y => y.EmailAdresses)));
    }
}

public class Customer
{
    public string CustomerId { get; set; } = default!;
    public ContactMethods ContactMethods { get; set; } = default!;
}

public class ContactMethods
{
    public string HomePhone { get; set; } = default!;
    public ICollection<Email> EmailAdresses { get; set; } = default!;
}

public class Email
{
    public string Address { get; set; } = default!;
    public string Type { get; set; } = default!;
}
```

This is the Exception which is thrown on `await context.BulkInsertOrUpdateAsync(entities, bulkConfig);`

```
System.InvalidOperationException
  HResult=0x80131509
  Message=Column 'ContactMethodsCustomerId' does not allow DBNull.Value.
  Source=Microsoft.Data.SqlClient
  StackTrace:
[...]

PR #1350, opened 2023-12-11
Research and possible start of a fix for issue 1337

fretje:
This includes a test for issue #1337 

And possible part of a fix, but I stumbled upon another roadblock.

With the current fix, I was able to get the foreign keys of the `Email` entities filled properly with their respective custId, and even inserted/merged into the database, but then an exception is thrown when trying to read the output from that operation, as `DbContext.Set<Email>().FromSqlRaw()` is used for that, and `DbContext.Set<TEntity>` isn't supported for owned types.

Not sure how to tackle this... but I thought I'd throw this out there already in case someone else has ideas?

### DIFF
--- a/EFCore.BulkExtensions.Tests/IncludeGraph/Issue1337.cs
+++ b/EFCore.BulkExtensions.Tests/IncludeGraph/Issue1337.cs
@@ -1,0 +1,70 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Xunit;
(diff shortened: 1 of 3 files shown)


borisdj:
Thx for contrib, don't have a solution out of the head, but will look into it.
If you managed to figure out a fix sooner, update the PR.

Karinza38 (approved):

borisdj:
@fretje Will add this now into the man branch, I will copy the code since PR has confilicts, or if you want to make new clean PR.

PR #1355, opened 2023-12-12
Add support for nested owned types

### LINKED ISSUE #1351: Owned types are only handled one level deep
When you have an object graph with owned types that go more than one level deep, only the first level is filled out. See the following test, which fails on the last line, because `SecondNestedProperty` is `null`.

```csharp
public class NestedOwned
{
    [Fact]
    public async Task NestedOwnedTest()
    {
        using var context = new NestedDbContext(ContextUtil.GetOptions<NestedDbContext>(databaseName: $"{nameof(EFCoreBulkTest)}_NestedOwned"));

        context.Database.EnsureDeleted();
        context.Database.EnsureCreated();

        NestedRoot[] entities = [
            new()
            {
                NestedRootId = "nestedrootid",
                FirstNested = new()
                {
                    FirstNestedProperty = "firstnested",
                    SecondNested = new()
                    {
                        SecondNestedProperty = "secondnested"
                    },
                },
            }
        ];

        await context.BulkInsertOrUpdateAsync(entities);

        var nestedroot = await context.Set<NestedRoot>().SingleAsync();
        Assert.Equal("nestedrootid", nestedroot.NestedRootId);
        Assert.Equal("firstnested", nestedroot.FirstNested.FirstNestedProperty);
        Assert.Equal("secondnested", nestedroot.FirstNested.SecondNested.SecondNestedProperty); // <-- this fails
    }
}

public class NestedRoot
{
    public string NestedRootId { get; set; } = default!;
    public FirstNested FirstNested { get; set; } = default!;
}

public class FirstNested
{
    public string? FirstNestedProperty { get; set; }
    public SecondNested SecondNested { get; set; } = default!;
}

public class SecondNested
{
    public string? SecondNestedProperty { get; set; }
}

public class NestedDbContext(DbContextOptions opts) : DbContext(opts)
{
    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<NestedRoot>(r => r.OwnsOne(r => r.FirstNested, f => f.OwnsOne(f => f.SecondNested)));
    }
}
```

fretje:
Fixes #1351

Owned types are now recursively checked for nested owned types.

This "recursive check" had to happen in 3 places. Once in TableInfo, and 2 times in SqlServerAdaptor.
I have implemented it using a local function to keep the code more or less in the same shape.

This is now only fully implemented for sql server. The other adaptors will also need to be updated to take advantage of this (but they should keep working like before without any changes).

From debugging and implementing this, it feels like there's probably quite some code in SqlServerAdaptor that can be removed as the properties on tableInfo can be used for a lot of the things that are "re-calculated" there now, I suppose? (Probably leftovers from previous refactors, that haven't made it to the SqlServerAdaptor?)

Anyways, bear with me. I've only recently found this repository, so I'm not too familiar yet with the code. It is a bit daunting at first to be fair (some of the long methods in TableInfo and the different adapters probably wouldn't mind some refactoring into smaller pieces ;-)).

So please review thouroughly.

Btw, I really like the functionality this library is providing! Right now, I'm using (and have also contributed to) SqlBulkTools for doing bulk imports, but this library seems much more promising with more features... at least if we get these bugs squashed out...

### DIFF
--- a/EFCore.BulkExtensions/TableInfo.cs
+++ b/EFCore.BulkExtensions/TableInfo.cs
@@ -556,73 +556,86 @@ public class TableInfo
 
             if (HasOwnedTypes)  // Support owned entity property update. TODO: Optimize
             {
-                var ownedTypes = OwnedRegularTypesDict.Values.ToList();
-                foreach (var navigationProperty in ownedTypes)
+                foreach (var navigationProperty in OwnedRegularTypesDict.Values.ToList())
                 {
-                    var property = navigationProperty.PropertyInfo;
-                    FastPropertyDict.Add(property!.Name, FastProperty.GetOrCreate(property));
+                    AddOwnedType(navigationProperty);
 
-                    // If the OwnedType is mapped to the separate table, don't try merge it into its owner
-                    if (OwnedTypeUtil.IsOwnedInSameTableAsOwner(navigationProperty) == false)
-                        continue;
-
-                    //Type navOwnedType = type?.Assembly.GetType(property.PropertyType.FullName!) ?? throw new ArgumentException("Unable to determine Type"); // was not used
-                    var ownedEntityType = context.Model.FindEntityType(property.PropertyType);
-                    if (ownedEntityType == null) // when entity has more then one ownedType (e.g. Address HomeAddress, Address WorkAddress) or one ownedType is in multiple Entities like Audit is usually.
+                    void AddOwnedType(INavigation navigationProperty, string prefix = "")
                     {
-                        ownedEntityType = context.Model.GetEntityTypes().SingleOrDefault(x => x.ClrType == property.PropertyType && x.Name.StartsWith(entityType.Name + "." + property.Name + "#"));
-                    }
-                    var ownedEntityProperties = ownedEntityType?.GetProperties().ToList() ?? new();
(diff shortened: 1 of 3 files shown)


EFCore.BulkExtensions/TableInfo.cs
```diff
@@ -556,74 +556,87 @@ public void LoadData<T>(DbContext context, Type? type, IEnumerable<T> entities,
...
+                                    FastPropertyDict.Add(ownedEntityPropertyFullName, FastProperty.GetOrCreate(ownedEntityProperty.PropertyInfo));
+                                }
                             }
-                            if (AreSpecifiedPropertiesToExclude && (BulkConfig.PropertiesToExclude?.Contains(ownedPropertyFullName) ?? false))
+
+                            var converter = ownedEntityProperty.GetValueConverter();
+                            if (converter != null)
                             {
-                                doAddProperty = false;
+                                ConvertibleColumnConverterDict.Add($"{prefix}{ownedEntityProperty.Name}", converter);
```
fretje, line 603: Here I'm not 100% sure this is correct. this used to be `ownedProperty.Name`, while prefix contains `property.Name`.

borisdj:
Great work, will check it soon.
I do have a ToDo list for refactoring, just to find the time.

PR #1357, opened 2023-12-16
PostgreSqlAdapter, correct columnType for decimal[]

gvivictor:
BulkInsert (PostgreSql) row with decimal[] column throws: System.InvalidCastException: 'Writing values of 'System.Decimal[]' is not supported for parameters having DataTypeName 'numeric'.'
Entity type to reproduce:
```
public class Row
{
    public int Id { get; set; }
    public required decimal[] Values { get; set; }
}
```


borisdj:
Thx for contrib.

PR #1428, opened 2024-03-04
fix(postgre): correctly count number of constraints

MaximBalaganskiy:
When multiple fields are provided the existing query gives the count which is greater than 1. This later causes a new constraint creation. This in its' turn cannot be run in concurrent environment.


borisdj:
Thx for contrib.

ivanbiljan:
Hey @borisdj , any estimates on when we can expect a nuget release?

borisdj:
8.0.3 published

PR #1435, opened 2024-03-15
Correct bulk operation on Sqlite Geometry types

### LINKED ISSUE #1203: Sqlite spatialite seems to be able to map only towards a Geometry but not the rest of the types
Hi, I am getting the following exception using Sqlite spatialite module:

System.InvalidOperationException: 'No mapping exists from object type NetTopologySuite.Geometries.LineString to a known managed provider native type.'

If I don't use the EFCore.BulkExtension the add and savechanges is working as expected. Using BulkInsert I am getting the above exception. Is the implementation missing or is there any work around to add a mapping for LineStrings and Points?

Thank you

### LINKED ISSUE #1214: Update to 7.1.5 EFCore.BulkExtensions.Sqlite and still the same problem with Geometries
Hi, I still have the same issue with the LineString and Points in SQLite when bulk inserting. Getting the error I reported earlier this month about not being able to map LineString (and Points) to any know Geomerty type. Has the changes you put being through this 7.1.5 version nuget or are they still pending?

Thank you

(sorry I couldn't find the original and I posted a new one but I have now update the original issue 1203

kvpt:
Hi,

Issue 1023 and 1214 are mentioned as fixed but it doesn't seems to be the case.

When inserting a point type with bulk insert the following exception is throw
>  No mapping exists from object type NetTopologySuite.Geometries.Point to a known managed provider native type.

This is because only Geometry and LineString types are handled by the [adapter](https://github.com/kvpt/EFCore.BulkExtensions/blob/7de0593127f779c6e3a928b6c7f74fa56f7457bf/EFCore.BulkExtensions/SqlAdapters/Sqlite/SqliteAdapter.cs#L495).

A more robust approach is to use instead the Geometry type only and rely on polymorphism.
However this alone doesn't work, the error is now
> '<GeoColumn> violates Geometry constraint [geom-type or SRID not allowed]'

To solve this we have two things to do :
- The first is to the right ordinate in the writer, I reused the code from EfCore [here](https://github.com/dotnet/efcore/blob/c3d3dc69f20c91e5f31aac1a126b375f8fccc5cf/src/EFCore.Sqlite.NTS/Storage/Internal/SqliteGeometryTypeMapping.cs#L131).
- The second is to use the same SRID value between the one declared in the context and the one of the geometry value
The previous code used the SRID defined in the BulkOperation configuration but I don't think it's the right think to do here because if the SRID is not defined in the context (and so in the database) the SRID need to be 0 and not the SRID from the BulkOperation. Another issue is that in an entity two geometry column can have different SRID is highly unlikely but technically possible.
The right thing to do is either to do nothing and only use the one provided at the geometry creation or to get the one defined in the Context. I tried to do this by using the GetSrid method on the property but I get nothing.
So I choose to fallback on the "passthrough" way and ignore the value defined in the BulkOperation configuration.

Fix #1203
Fix #1214

### DIFF
--- a/EFCore.BulkExtensions.Tests/EFCoreBulkTestAtypical.cs
+++ b/EFCore.BulkExtensions.Tests/EFCoreBulkTestAtypical.cs
@@ -853,7 +853,8 @@ public class EFCoreBulkTestAtypical
                     Street = "Some Street nn",
                     LocationGeography = new Point(52, 13),
                     LocationGeometry = new Point(52, 13),
-                    GeoLine = new LineString(new List<Coordinate> { new Coordinate(52, 13), new Coordinate(50, 12) }.ToArray())
+                    GeoLine = new LineString(new List<Coordinate> { new Coordinate(52, 13), new Coordinate(50, 12) }.ToArray()) { SRID = 4326 },
+                    GeoPoint = new Point(52, 13) { SRID = 4326 }
                 }
             };
 
--- a/EFCore.BulkExtensions.Tests/TestContext.cs
+++ b/EFCore.BulkExtensions.Tests/TestContext.cs
@@ -203,7 +203,13 @@ public class TestContext : DbContext
 
             modelBuilder.Entity<ItemHistory>().ToTable(nameof(ItemHistory));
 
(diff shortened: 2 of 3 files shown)


borisdj:
Thx for contrib.
New version will be published in a week day with new nuget.

PR #1445, opened 2024-03-18
Remove unneeded dependency

kvpt:
SqlClient is a SqlServer specific dependency remove it from others providers


borisdj:
Not so simple.
There are few dependencies in code.
For example
`SqlBulkCopyOptions`
`public override object CreateParameter(SqlParameter sqlParameter)`
and first would need to check if those can be refactored to removed the need for SqlClient.

kvpt:
You're right, I just checked that the project compiled but now I see the content of the csproj there are just dummy files so it's not a guarantee.

borisdj:
When compiling for per Adapter, source files needs to be copied into specific folders.
There is a script for that in the project.

borisdj:
At the moment only MySql adapter does not have this dependency.

borisdj:
Thx for contrib.

PR #1446, opened 2024-03-18
Make MySql adapter respect UseTempDb flag when creating table copies

### LINKED ISSUE #1434: MySQL adapter ignores UseTempDb flag
What exactly is the rationale behind this?
image

 I would prefer not having to give CREATE permissions unless absolutely necessary.

ivanbiljan:
Closes #1434


borisdj:
Thx for contrib.

PR #1518, opened 2024-06-16
Library creates uninformative exception when UpdateByProperties list has wrong (enexisting) property names

TrueGandalf:
allProperties.First(p => p.Name == b).GetColumnName(ObjectIdentifier) throws uninformative no elements exception when bulkConfig has an unexisting propertyName in the UpdateByProperties list


borisdj:
Thx for contrib

Issue #669, opened 2021-11-29
SQL Server/PostgreSQL conflict: ambiguous reference for IndexBuilder


echernyavsky:
Starting from version 6.0.4 there is an issue with IncludeProperties method in IndexBuilder.
```
Ambiguous invocation: Microsoft.EntityFrameworkCore.Metadata.Builders.IndexBuilder<TEntity> IncludeProperties<TEntity>(this Microsoft.EntityFrameworkCore.Metadata.Builders.IndexBuilder<TEntity>, System.Linq.Expressions.Expression<System.Func<TEntity,object?>>) (in class SqlServerIndexBuilderExtensions) Microsoft.EntityFrameworkCore.Metadata.Builders.IndexBuilder<TEntity> IncludeProperties<TEntity>(this Microsoft.EntityFrameworkCore.Metadata.Builders.IndexBuilder<TEntity>, System.Linq.Expressions.Expression<System.Func<TEntity,object>>) (in class NpgsqlIndexBuilderExtensions) match
```

The problem is that the EFCore.BulkExtensions now refers Microsoft.EntityFrameworkCore.SqlServer and Microsoft.EntityFrameworkCore.PostgreSQL which has completely the same set of extensions for EF builders. 

Looks like it will be better to create separate packages for Postgres and SQL Server implementations.

borisdj:
How do you reproduce this problem?
In RunUpdate test it works without issue:
`PropertiesToInclude = new List<string> { nameof(Item.Description) },`

echernyavsky:
```
public class MyEntityConfig : IEntityTypeConfiguration<MyEntity>
{
    protected override void ConfigureEntity(EntityTypeBuilder<MyEntity> builder)
    {
        builder.HasIndex(it => it.UserId)
            .IncludeProperties(it => new { it.ReadAt, it.DeletedDate });
    }
}
```

The problem is that even if I specify the extensions class implicitly there is an autogenerated snapshot that have the same issue.

borisdj:
Try this fix at the moment:
https://github.com/dotnet/efcore/issues/20780

borisdj:
Use explicit call like this:
```C#
protected override void ConfigureEntity(EntityTypeBuilder<MyEntity> builder)
{
	builder.HasIndex(it => it.UserId);
	Expression<Func<MyEntity, object?>> expr = it => new { it.DeletedDate };
[...]

PR #1537, opened 2024-07-26
Introduce a core package which contains only shared code

kvpt:
Hi,

This Pull Request introduce a dedicated package to contains shared code, the use case is as follow :
> I have a library that do BulkOperation, and I want the library to be provider agnostic and be able to provider the provider in the project referencing the library.

With the current packages I can't do that, I tried to reference the full package and mark it as private dependency and use the provider specific package in the application project. But this not working because the assemblies are different (perhaps possible to fix that by using type forwarder or assembly redirection, but I don't want to have to do this on each application).

So for me the solution is to introduce a generic package in addition of the provider specific one.
Efcore do this, Microsoft.EntityFrameworkCore.Abstractions and Microsoft.EntityFrameworkCore are generic then we have Microsoft.EntityFrameworkCore.SqlServer for example for SqlServer.

Doing this way mean that EFCore.BulkExtensions will now contains only the abstraction/shared part, and provider projects will need to depend on it. This is the more natural way of doing that but the issue is that it will broke all consumers projects after update by forcing them to add the provider package that they use.
We can introduce it on major version to limit the impact but to do that this late (in term of library lifetime) seems risky to me.

So I go to the other way of doing that, introduce a specific package to contains the generic code, the Core package.
So now each provider project reference the core package and the global package is now a meta package that only reference all providers package.

So now in a library project I can only reference the core package and the provider specific in the application project.

The only question I ask is about the EFCore.BulkExtensions nuget package structure.
Either having a "pure" meta package with only dependencies like I have done or having a package with all the source code copied into it and no dependencies to providers packages like now.
For the consumer it will work the same, it just a preference of the package author.

I also introduced a prop file to centralize project properties to facilitate the maintainability.
And as the properties are now centralized the bat scripts are now useless so I deleted them.
Now to increment the version just change it in the prop file at the project root, and compiling the code will place all nupkg in the nuget folder automatically.


borisdj:
Will need some time to analyse and consider this.
I had a version similar to this, with base project and sufixes .**Core** and .**Common**, more info here:
https://github.com/borisdj/EFCore.BulkExtensions/issues/669#issuecomment-1340884834
But later reverted  to current structure, to keep it simple and backwards compatible.

kvpt:
With the code copy way it can produce conflict if the package overlap, if the global package and the provider specific are used at the same time (Not sure it can't be prevented at compile time).
With the project reference way this problem can't happen, the package management layer will handle that.
But by doing this way to preserve backwards compatibility we have to introduce a core/common specific project instead of using the global package EFCore.BulkExtensions to that role. That's what I did.

borisdj:
If I understand correctly this structure requires core project to have its own new nuget, and current packages, main and specific ones, would have that dependency.

kvpt:
Exactly.

Having the core package as a dependency is mandatory, if we want to have a library project with only the core dependency and an application with the full package or one or several provider specific project without causing conflict.

We can make the full package without dependency like now (by copy the source code like before) but in this case the full package will not be usable with core or provider package, this will cause conflict.
And as we can't control projects transitives dependencies I dont think it is a practical thing to do.

Like I said the cleaner way to do this would be that EFCore.BulkExtension is the core package instead of the full package but this will introduce a breaking change for consumer of the package.
So the most transparent way to do this is to have a core/common package as the base for all others.

kvpt:
Note that for current users of the packages, the core package will not be required explicitly and will be automatically imported as transitive dependency.

borisdj:
Yes, I am aware that Core would be automatically installed as transitive dependency, which is a good thing here.
As for the use-cases I do not see a situation where someone needs main package and a specific one at the same time, since main contains all providers. But there was a question about using only 2 specific ones nuget, which currently is creating a conflict (only solution there was to use main one with all 4 provider dependencies). 
Now with this change it is possible to combine any 2 or 3 if needed.

kvpt:
Yes it is, as each parts have now a dedicated part so it's fully composable.

borisdj:
Big praises for this work.

borisdj:
Big praises and gratitude for this work.

PR #1542, opened 2024-08-10
Add inherited & owned logic

zyofeng:
Currently if an entity is owned by multiple parent and one of them happen to be inherited, the following line will not be able to find ownedEntityType because entityType is the inherited class instead of base class
```
if (ownedEntityType == null) // when entity has more then one ownedType (e.g. Address HomeAddress, Address WorkAddress) or one ownedType is in multiple Entities like Audit is usually.
                        {
                            ownedEntityType = context.Model.GetEntityTypes().SingleOrDefault(x => x.ClrType == property.PropertyType && x.Name.StartsWith(entityType.Name + "." + property.Name + "#"));
                        }
```

### DIFF
--- a/EFCore.BulkExtensions.Core/TableInfo.cs
+++ b/EFCore.BulkExtensions.Core/TableInfo.cs
@@ -586,11 +586,7 @@ public class TableInfo
                         prefix += $"{property.Name}_";
 
                         //Type navOwnedType = type?.Assembly.GetType(property.PropertyType.FullName!) ?? throw new ArgumentException("Unable to determine Type"); // was not used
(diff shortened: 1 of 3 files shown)


borisdj:
Thx for contrib.
Btw I have made a change to keep this fix but also added condition to use prior code segment when Ownership return more then one element, not Single but multiple. 
Test for this is `[InlineData(SqlType.SqlServer)] private void OwnedTypesTest(SqlType sqlType))`

zyofeng:
thanks, would you mind making a release for this? I've been using a forked version in prod currently

Issue #1554, opened 2024-08-26
BulkUpdateAsync throws KeyNotFoundException on OwnsMany with JSONB column


AndriyKhavro:
Here are the repro steps: https://github.com/AndriyKhavro/BulkExtensionJsonb
I followed [this article](https://medium.com/@serhiikokhan/jsonb-in-postgresql-with-ef-core-cc945f1aba2a) to map a collection property of an entity to JSONB type. Calling `BulkUpdateAsync` on this type throws KeyNotFoundException.

Here is the db context:
```
public class Product
{
    public int Id { get; set; }
    public string Name { get; set; } = string.Empty;
    public List<Review> Reviews { get; set; } = [];
    public DateTimeOffset CreatedAt { get; set; } = DateTimeOffset.UtcNow;
    public DateTimeOffset UpdatedAt { get; set; } = DateTimeOffset.UtcNow;
}

public class Review
{
    public string User { get; set; }
    public string Content { get; set; }
    public int Rating { get; set; }
}


public class ProductContext : DbContext
{
    public DbSet<Product> Products => Set<Product>();

    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        => optionsBuilder
            .UseNpgsql("Server=localhost;Port=5432;Database=BulkExtensionJsonb;Username=postgres;Password=somepassword;");

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder
            .Entity<Product>()
            .OwnsMany(product => product.Reviews, builder => { builder.ToJson(); });
    }
}
```

And the code that throws the exception:
```
﻿using BulkExtensionJsonb;
using EFCore.BulkExtensions;

int id = Random.Shared.Next();

var context = new ProductContext();
await context.Products.AddAsync(new Product { Id = id });
await context.SaveChangesAsync();

var product = await context.Products.FindAsync(id);

product!.UpdatedAt = DateTimeOffset.UtcNow;
await context.BulkUpdateAsync(new[] { product }); // throws System.Collections.Generic.KeyNotFoundException
```

Call stack:
```
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'Reviews' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
[...]

PR #1633, opened 2024-12-09
Added support for Npgsql json owned types

jeremy-morren:
Adds support for Postgresql owned types mapped as json.

Also added Postgresql to the databases tested for CompositeKeys (works without any changes).

Fixes https://github.com/borisdj/EFCore.BulkExtensions/issues/1554


borisdj:
Thx for Contrib.

PR #1653, opened 2025-01-14
Add support to set ConflictOption in MySqlBulkLoader

buecom:
The MySqlBulkLoader class has a field ConflictOption to specify how to treat duplicates during bulk inserts. Possible values are None, which is the default and current behavior in this project, Ignore and Replace.
This PR lets the user specify the wished for behavior via a new BulkConfig field, which is used when a new MySqlBulkCopy instance gets configured. If nothing is explicitly specified, the behavior is the same as the current one.


borisdj:
Thx for Contrib.

PR #1662, opened 2025-01-22
Fixed owned type value converter

jeremy-morren:
Fixed a bug where owned types did not use type mapping converter.


borisdj:
Thx for contrib.

PR #1663, opened 2025-01-22
Updated ContextUtil configuration updates to DI

jeremy-morren:
Updated ContextUtil to use SqlType as a constructor, rather than a static property.


borisdj:
Hi, I have merged some other PRs so now there is a conflict, could you resolve it, so that can finish the Merge.
And Thx for Contrib.

jeremy-morren:
Commit updated.

To get the project to compile with Oracle, it was necessary to update provider packages off of prerelease (except for MySQL, where latest package is still a prerelease).

PR #1680, opened 2025-02-18
Provide server more dynamically.

### LINKED ISSUE #1679: Choose adapter manually?
Hi,

I am running tests against multiple databases. But the database adapter seems to be cached in static fields and therefore, when I run all tests, the wrong adapter is picked.

Is there an option to define the adapter manually or inject into the dbcontext or something like that. I have one context per database.

SebastianStehle:
Closes #1679 

I tried to implement it with as few changes as possible. I think there are too much calls to the SqlAdaptersMapping and I also do not understand why sometimes the adapters resolve the server, but it is not the goal of the PR to "improve" that.

Some tests are failing but I could not find the root cause yet, perhaps they are just failing in general.

### DIFF
--- a/EFCore.BulkExtensions.Core/Batch/BatchUpdateCreateBodyData.cs
+++ b/EFCore.BulkExtensions.Core/Batch/BatchUpdateCreateBodyData.cs
@@ -38,7 +38,7 @@ public class BatchUpdateCreateBodyData
         LambdaExpression updateExpression)
     {
         BaseSql = baseSql;
-        DatabaseType = SqlAdaptersMapping.GetDatabaseType();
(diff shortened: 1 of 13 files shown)


SebastianStehle:
Can I support you to get this out? btw: I have written you an email a while ago about licensing.

borisdj:
You can check the tests that are failing and try locate the issue.

SebastianStehle:
Sure. Do you have a docker compose or something to setup all the test dbs?

borisdj:
No, I have all Database servers installed, and tests (with xunit) when run are using EF to create Db if it does not exist.

PR #1684, opened 2025-02-22
Use a better way to check if a field current value is the default value

kvpt:
The current code used to check if a field value is equals to the default value of his type is wrong.
The compare of the toString method return doesn't work on custom struct (like LocalDate type in the test project).
Also the checks and the order of them are not optimal.
This PR introduce a more correct and robust way to do that.


borisdj:
Thx for Contrib.

PR #1728, opened 2025-04-08
Check UseTempDb in Postgres v8

SuperTroye:
I see this was fixed in v9, but not in v8. Hoping this can be added in since my project is still in Dotnet 8


borisdj:
Sure, thx for Contrib.

SuperTroye:
> Sure, thx for Contrib.

Thanks! Can you also send to Nuget (8.1.4?) since that's where we get the library.

borisdj:
@SuperTroye new versions of Nugets, both v9 and v8, will be published next week.

PR #1740, opened 2025-04-17
Replace constraint name check with count verification in MysqlAdapter

hernandevelop:
- Change SELECT from CONSTRAINT_NAME to COUNT(*) in SQL query
- Fixes bug where existence check relied on non-unique constraint name retrieval
- Ensures deterministic verification of unique constraint existence
- Addresses concurrency issues in constraint creation scenarios


borisdj:
Thx for Contrib.

PR #1741, opened 2025-04-18
Correct unique index verification query

hernandevelop:
- Implement JOIN between STATISTICS and TABLE_CONSTRAINTS  
- Verify exact column sequence match via GROUP_CONCAT(COLUMN_NAME)
- Wrap in COUNT(*) subquery for deterministic existence checking


hernandevelop:
When using BulkConfig, you must explicitly define the CustomDestinationTableName to exactly match an existing unique constraint:

```
new BulkConfig {
    CustomDestinationTableName = "{schema}.{table}",   // ← important!  
    UpdateByProperties = new List<string> { "Column1", "Column2" }
};
```

PR #1746, opened 2025-04-30
Apply topological sort on delete to .NET >= 8.0.

filmstarr:
This change applys the topological sort on deletions to .NET 9 (and future higher versions), as well as .NET 8.


borisdj:
Thx for Contrib.

PR #1747, opened 2025-05-20
Introduce bulk context

SebastianStehle:
The Server interface was problematic:

1. There were calls to the AdapterMapping all over the place.
2. The IServer interface had a property for connection and transaction, even though they do not have a singleton context and could cause concurrency issues.

Therefore I introduced a new BulkContext class which is a wrapper over DbContext and IServer and is passed through everywhere. It makes the actual code faster and more stable and easier to understand in my opinion.


borisdj:
Thx for the PR.
Will check it in the following days.

borisdj:
Merged and new nuget published.

PR #1762, opened 2025-06-18
Fix #1430: Enable BulkRead on Database Views

### LINKED ISSUE #1430: BulkRead does not work for views
Hi, when I try to use `BulkRead` to read entities from a SQL Server view, the method throws an exception 'Entity does not contain a table name'. When I look into the code, I think the `IEntityType.GetTableName()` form Entity Framework Core is used. This method indeed will return `null` for entities mapped to SQL views. `IEntityType.GetViewName()` should probably be used instead.

Is it possible to add support to BulkExtensions.BulkRead to read entities from a view? That would be great!

Versions being used by me at the moment:
- Microsoft.EntityFrameworkCore version 7.0.11
- EFCore.BulkExtensions version 7.1.6

entity definition:
```
public class MyViewItem
{
    public long Id { get; set; }
    
    public int TenantId { get; set; }
    
    [Required]
    [MaxLength(100)]
    public string Description { get; set; }
}
```
DbContext:
```
public class MyContext : DbContext
{
    public DbSet<MyViewItem> MyViewItems { get; set; }
    
    public MyContext(DbContextOptions<MyContext> options) : base(options)
    {
    }

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<MyViewItem>(entity =>
        {
            entity.HasKey(item => new { item.TenantId, item.Id });
            entity.ToView("MyView");
        });
    }
}
```

Code to read (which throws the exception):
```
public async Task<IEnumerable<MyViewItem>> ReadViewItems(List<long> ids)
{
    var tenantId = 1;
    var entities = ids
        .Select(id => new MyViewItem { Id = id, TenantId = tenantId })
        .ToList();

    using (var context = _contextFactory.Create())
    {
        await context.BulkReadAsync(entities, config =>
        {
            config.UpdateByProperties = new List<string> { nameof(MyViewItem.TenantId), nameof(MyViewItem.Id) };
        }).ConfigureAwait(false);

        return entities;
    }
}
```

SQL Server view
````
CREATE OR ALTER VIEW [MyView] AS
SELECT [TenantId]
     , [Id]
     , [Description]
FROM   [MyTable]
````

m0ddixx:
I propose this pull request to fix the issue #1430 to enable BulkRead on database views.
I have opted for a minimal approach. For review, I'd also suggest to look into customTableName since we could also overwrite it.

### DIFF
--- a/EFCore.BulkExtensions.Core/TableInfo.cs
+++ b/EFCore.BulkExtensions.Core/TableInfo.cs
@@ -195,11 +195,15 @@ public class TableInfo
             }
         }
 
-        Schema = customTableName != null ? customSchema : entityType.GetSchema() ?? defaultSchema;
-
         var entityTableName = entityType.GetTableName();
         var entityViewName = entityType.GetViewName();
         bool isView = entityTableName == null && entityViewName != null;
+        
+        if(isView)
+            Schema = customTableName != null ? customSchema : entityType.GetViewSchema() ?? defaultSchema;
+        else
+            Schema = customTableName != null ? customSchema : entityType.GetSchema() ?? defaultSchema;
+
         TableName = customTableName ?? (isView ? entityViewName : entityTableName);
 
         string? sourceSchema = null;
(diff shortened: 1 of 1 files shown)


borisdj:
Thx for Contrib

Issue #1753, opened 2025-06-10
Bug Report: EFCore BulkExtensions Cannot Alter Timestamp Column


tunahanaliozturk:
# 🐛 Bug Report: EFCore BulkExtensions Cannot Alter Timestamp Column

## **Problem Description**
The synchronization process fails when attempting to bulk sync `ARImgEntity` data from cloud to local database with the following error:

```
Cannot alter column 'Ts' to be data type timestamp.
```

## **Error Details**
- **Operation**: `FetchFromCloud("arimg/fetch", db.ARImg, "arImgArray", BulkSyncType.InsertOrUpdateOrDelete)`
- **Error Location**: EFCore.BulkExtensions temp table column alteration
- **Specific Error**: `ALTER TABLE [dbo].[ARImgTempa8b256ebOutput] ALTER COLUMN [Ts] rowversion;`

## **Root Cause Analysis**
The issue occurs because:
1. The `Ts` property is configured as `rowversion`/`timestamp` in Entity Framework
2. EFCore.BulkExtensions attempts to create a temporary table and alter the `Ts` column type
3. SQL Server **does not allow** altering existing columns to `timestamp`/`rowversion` data type
4. The current exclusion logic only excludes `byte[]` properties named "Ts", but the BulkExtensions library still tries to process the column

### Entity Configuration

// ARImgEntity.cs
[Timestamp]
public byte[]? Ts { get; set; }

// ARImageEntityConfiguration.cs
builder.Property(e => e.Ts)
      .IsRowVersion()
      .IsConcurrencyToken();
```

### Current Exclusion Logic

// SynchronizationService.cs - FetchFromCloud method
var excludedProperties = typeof(T).GetProperties()
  .Where(p => p.Name == "Ts" && (
      p.PropertyType == typeof(byte[])
  ))
  .Select(p => p.Name)
  .ToList();

if (excludedProperties.Count != 0)
    bulkConfig.PropertiesToExclude = excludedProperties;
```

### Option 1: Enhanced Property Exclusion (Recommended)
```csharp
var excludedProperties = typeof(T).GetProperties()
    .Where(p => p.Name == "Ts" && (
        p.PropertyType == typeof(byte[]) || 
        p.PropertyType == typeof(byte[]?)
    ))
    .Select(p => p.Name)
    .ToList();

// Additionally exclude rowversion/timestamp properties by attribute
var timestampProperties = typeof(T).GetProperties()
[...]

PR #1765, opened 2025-06-24
Add AutoExcludeTimestamp support to BulkConfig

Moongazing:
When enabled, this flag auto-detects [Timestamp] / .IsRowVersion() columns and excludes them from all relevant insert, update, and compare operations by automatically populating PropertiesToExclude.

This prevents SqlBulkCopy errors caused by rowversion columns, especially in entities using concurrency tokens.

    Fully backwards-compatible

    Opt-in via BulkConfig.AutoExcludeTimestamp = true

     Works alongside IgnoreRowVersion and other property filters

Closes issue #1753


borisdj:
Thx for contrib.

PR #1877, opened 2025-10-28
Problems with MySQL BulkUpdate Operations

muratalicilar:
```
1. TableInfo/LoadData -> PropertyColumnNamesUpdateDict
2. MySqlQueryBuilder/MergeTable -> PropertyColumnNamesUpdateDict
```

The problem I'm experiencing:

When performing a bulk update, I only pass the fields I want to change into the "`PropertiesToInclude`" field.

However, because the main table had multiple unique fields, and these fields were not included in the "`PropertiesToInclude`" field, the records couldn't be inserted into the temporary table. Because the unique fields weren't included in the "`PropertiesToInclude`" field, they were being passed to the temporary table with a "`null`" value.

I don't want the unique fields to be changed again.

When I reviewed the project, I realized that the "`PropertiesToIncludeOnUpdate`" field was not functional, so I injected it into the relevant fields, and now everything works as I want.

In "`PropertiesToInclude`," I add the fields that need to be passed to the temporary table and that need to be taken action (the fields that need to be changed and the unique fields).
In "`PropertiesToIncludeOnUpdate`," I add only the fields I want to change.
In "`PropertiesToExcludeOnUpdate`," I add the fields I don't want to change.

Thank you.

### DIFF
--- a/EFCore.BulkExtensions.Core/TableInfo.cs
+++ b/EFCore.BulkExtensions.Core/TableInfo.cs
@@ -659,12 +659,23 @@ public class TableInfo
                                     doAddProperty = false;
                                 }
 
+                                bool doAddPropertyOnUpdate = true;
+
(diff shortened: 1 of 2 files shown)


borisdj:
Hi, can you also make and add a Test for this.

PR #1891, opened 2025-11-12
Attempt to add .NET10 support

molinch:
The only code change is in IQueryableExtensions, the rest is targeting .NET10.
However, MySQL and Oracle don't have any support yet so it would be broken with these changes...
PostgreSQL has a rc2 but I'm also not sure of the impact of using it.

Maybe this should really be *just* a preview version just to unblock SQL Server/SQLite folks and potentially PostgreSQL folks.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,7 +7,7 @@
     <Company>CODIS LLC</Company>
     <PackageIcon>EFCoreBulk.png</PackageIcon>
     <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
-	<PackageReleaseNotes>net 9 full</PackageReleaseNotes>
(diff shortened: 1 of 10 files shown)


davidnmbond:
[The PanoramicData fork](https://github.com/panoramicdata/EFCore.BulkExtensions) has a pre-release nuget up [here](https://www.nuget.org/packages?q=EFCore.BulkExtensions.Dotnet10&includeComputedFrameworks=true&prerel=true&sortby=relevance) (EFCore.BulkExtensions.Dotnet10).

No Oracle or MySQL support yet, as you say.

borisdj:
I have made partial changes to some providers, published as RC are MsSql and Lite
v 10.0.0-rc.1

Issue #1780, opened 2025-07-31
ChildId property excluded from update after exception in BulkInsertOrUpdate with IncludeGraph=true on PostgreSQL


Lion-Ch:
I'm using the following packages:

```
EFCore.BulkExtensions.PostgreSql v8.1.1
Microsoft.EntityFrameworkCore v8.0.11
Npgsql.EntityFrameworkCore.PostgreSQL v8.0.11
```

I have a parent entity with a child entity, both with Id = 0 before insert. I perform BulkInsertOrUpdate with IncludeGraph = true. On the first successful insert, both entities are saved correctly and the generated IDs are returned as expected.

However, if an exception occurs during the bulk operation (e.g., due to a constraint violation), then on all subsequent calls, even with a new DbContext instance, the ChildId property on the parent entity is always null.

From EF Core logs, I see that the generated SQL contains:

`DO UPDATE SET "ChildId" = EXCLUDED."ChildId"`

But ChildId is excluded from the actual update (treated as EXCLUDED), which means it is not updated properly.

I suspect the issue is related to internal caching of the entity/table metadata (TableInfo) inside EFCore.BulkExtensions. After an exception, the cache seems corrupted or incomplete, causing ChildId to be excluded from the update statement.

Notably, restarting the application does not solve the issue, which indicates the cache is static or persisted somewhere unexpected.

Lion-Ch:
A similar [question](https://stackoverflow.com/questions/61398094/efcore-bulkinsert-with-one-to-one-relationship?utm_source=chatgpt.com) on stackoverflow

PR #1901, opened 2025-11-25
[Issue #1780] Bug: IncludeGraph fails when first entity has null navigation property

Lion-Ch:
Fixed the case of inserting child entities, when there could be an entity with an empty child entity at the beginning of the array. Because of this, the subsequent insertion of bundles of entities was performed incorrectly.  `GroupBy` by `EntityType` in `ExecuteWithGraphAsync` put the type of entity that was the parent in the first place.

# Example

The main entity is **Parent - P**, the child entity is **Child - C**. 
If there was an array Parent[]
```csharp
[
new () { Child = null },
new () { Child = new() },
new () { Child = new() },
// ...
]
```
After sorting, we get: **PCPCP**

After grouping, we will insert in bundles: Parent[], Child[].  **Which is wrong!**

If all the Parents had Child, then after sorting we get: **CPCPCP**
After grouping, we will insert in bundles: Child[], Parent[]. That's right.

# Solution
To avoid this problem, an additional sorting by graph depth was added. 
For each entity, its depth (length to the vertex) is determined. The maximum depth is determined for each type of entity in order to understand which entities need to be inserted first (the type of entity with the maximum depth) and which last (the type of entity with the minimum depth).

1. For example, after sorting, we get: **PCPCP**
2. After determining the depth: P(0)C(1)P(0)C(1)P(0)
3. After determining the maximum depth for the type of entity: P = 0, C = 1
4. After applying depth sorting, we get: C(1)C(1)P(0)P(0)P(0)

Further, with GroupBy, the groups will be formed correctly and the identifiers for external entities will be entered correctly.

# In summary

- Added additional sorting of the graph by depth
- Added a test to check the insertion of entities (`BulkInsertOrUpdate_EntityWithNestedNullableObjectGraph_SavesGraphToDatabase1`)
- To activate the old logic, disable `withAdditionalSorting` in `GraphUtil`

# It remains to be done
- [ ] Roll back the database schema and write a new test
- [ ] Consider the cases I haven’t accounted for, because this error is the only one I have fixed.
- [ ] Add tests for more complex entity graphs and test the algorithm.
- [ ] Test performance on large amounts of inserted data (reflection, algorithmic complexity)

### DIFF
--- a/EFCore.BulkExtensions.Core/DbContextBulkTransactionGraphUtil.cs
+++ b/EFCore.BulkExtensions.Core/DbContextBulkTransactionGraphUtil.cs
@@ -40,7 +40,7 @@ internal static class DbContextBulkTransactionGraphUtil
         // If this is set to false, wont' be able to support some code first model types as EFCore uses shadow properties when a relationship's foreign keys arent explicitly defined
         bulkConfig.EnableShadowProperties = true;
 
-        var graphNodes = GraphUtil.GetTopologicallySortedGraph(context, entities);
+        var graphNodes = GraphUtil.GetSortedGraph(context, entities);
 
         if (graphNodes == null)
             return;
--- a/EFCore.BulkExtensions.Core/Util/GraphUtil.cs
+++ b/EFCore.BulkExtensions.Core/Util/GraphUtil.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
+
 using System;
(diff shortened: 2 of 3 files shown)


borisdj:
Merged, thx for Contrib.

Lion-Ch:
@borisdj  This merge request is incomplete (not all the tasks I listed at the end of the issue have been completed). I would recommend rolling it back and marking it as a draft.

borisdj:
@Lion-Ch 
I thing we can keep it, but will try to make upgrades soon, and fast fixes if any issues arise.
Also maybe you could make another PR to add make this SortingFeature configurabe in BulkConfig with default to False (for now, later maybe default could be True), but still to be able to switch of if not needed.

